PR #10, opened 2018-11-12
"actions can be used" test should use its action

Retsam:
Browsing through the `useObservable` tests, it seemed odd to me that the "actions can be used" test wasn't actually using its action to update the state. 

Apologies if I'm misunderstanding the purpose of the test.

### DIFF
--- a/test/useObservable.test.tsx
+++ b/test/useObservable.test.tsx
@@ -58,7 +58,7 @@ describe("is used to keep observable within component body", () => {
                 }
             })
             return (
-                <div onClick={() => obs.inc()}>
+                <div onClick={obs.inc}>
                     {obs.x}-{obs.y}
                 </div>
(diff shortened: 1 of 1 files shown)


FredyC:
Nice catch :) Although please change it to ` <div onClick={obs.inc}>`, there is no need for extra wrapping function since it's bound already.

Retsam:
Is it `this` bound already?  I didn't think `observable` did any autobinding, by default.  It's a moot point, either way, since it doesn't actually reference `this`, (so you're definitely right to switch it), I'm just curious.

FredyC:
Yea, there is `this` in play here, but action itself references the `obs` directly which is just fine. You could use `this` in there as well, but I feel it's like making a functional component dirty :)

Thanks for the fix.

PR #21, opened 2018-11-18
ref support for observer components

xaviergonz:


src/observer.ts
```diff
@@ -8,31 +8,41 @@ export function useStaticRendering(enable: boolean) {
...
-        // can be invalidated (see above) once a dependency changes
-        let rendering!: ReturnType<typeof baseComponent>
-        observerReaction.track(() => {
-            rendering = baseComponent(props)
+    const memoComponent = memo(
+        // we have to use forwardRef here because:
+        // 1. it cannot go before memo, only after it
+        // 2. forwardRef converts the function into an actual component, so we can't let the baseComponent do it
+        //    since it wouldn't be a callable function anymore
+        forwardRef((props: P, ref: any) => {
```
FredyC: I don't like the idea of having every observed component to be forwarding ref automatically. Can we like to figure how to do it optionally, eg. second argument to `observer` call? If not, then I would rather go with `forwardObserver` as a separate entity.

Official docs even mention it's not necessary that often and I personally have used it like once. It always fees like a hack if you need to communicate with a component directly.

> Ref forwarding is a technique for automatically passing a ref through a component to one of its children. This is typically not necessary for most components in the application.

xaviergonz: sounds good, done

README.md
```diff
@@ -14,10 +14,14 @@ Project is written in TypeScript and provides type safety out of the box. No Flo
...
 ## API documentation
 
-### `observer<P>(baseComponent: FunctionComponent<P>): FunctionComponent<P>`
+### `observer<P>(baseComponent: FunctionComponent<P>, options?: IObserverOptions): FunctionComponent<P>`
 
 Function that converts a function component into a reactive component, which tracks which observables are used automatically re-renders the component when one of these values changes. Observables can be passed through props, accessed from context or created locally with `useObservable`.
 
+As for options, it is an optional object with the following optional properties:
+
+-   `fowardRef`: pass `true` to use `forwardRef` over the inner component, pass `false` (the default) otherwise.
```
FredyC:
```suggestion
-   `forwardRef`: pass `true` to use `forwardRef` over the inner component, pass `false` (the default) otherwise.
```

src/ObserverComponent.ts
```diff
@@ -7,19 +7,20 @@ interface IObserverProps {
     render?(): ReactElement<any>
 }
 
-export const Observer = observer(({ children, render }: IObserverProps) => {
+function ObserverFunc({ children, render }: IObserverProps) {
```
FredyC:
```suggestion
function ObserverComponent({ children, render }: IObserverProps) {
```

src/observer.ts
```diff
@@ -8,31 +8,67 @@ export function useStaticRendering(enable: boolean) {
...
+    baseComponent: React.FunctionComponent<P>,
+    options?: IObserverOptions
+): React.NamedExoticComponent<P>
+
+// n.b. base case is not used for actual typings or exported in the typing files
+export function observer<P extends object, TRef = {}>(
+    baseComponent: React.RefForwardingComponent<TRef, P>,
+    options?: IObserverOptions
+) {
+    const realOptions = {
```
FredyC: Move this under the check for static rendering please.

xaviergonz: done

src/ObserverComponent.ts
xaviergonz: done

xaviergonz:
rebased

FredyC:
Some strange way for a rebase seeing the history you have caused here. Luckily there is a squash, so it's ok :)

xaviergonz:
Now you know why I usually merge! I always mess up rebases

FredyC:
```shell
git fetch
git rebase origin/master
git rebase --continue // if there were conflicts
git push --force
```

If you did something else in between, eg. pull (Git likes to recommend it after rebase), then you see a result :)

FredyC:
Released in [v0.3.0](https://github.com/mobxjs/mobx-react-lite/releases/tag/v0.3.0)

PR #26, opened 2018-11-18
Use observable effect disposer test

### LINKED ISSUE #25: Possible memory leak in useObservableEffect
Probably not, but I just noticed in test coverage that the disposer is never called.

```ts
    useEffect(
        () => () => {
            if (disposerRef.current) {
                disposerRef.current()
            }
        },
        inputs
    )
```

I tried calling `unmount`, modify the store value again and render counters were the same so, in the essence, it seems to be working. However, the coverage hasn't changed with such test so it never gets called either way it seems.

In theory, it could be caused by the [asynchronous nature of `useEffect`](https://github.com/kentcdodds/react-testing-library/issues/215), but I tried switching it to `useLayoutEffect` which should be synchronous, but nothing has changed.

I am not really sure what might be wrong here and how to test it properly.

@xaviergonz

xaviergonz:
Closes #25

### DIFF
--- a/test/useObservableEffect.test.tsx
+++ b/test/useObservableEffect.test.tsx
@@ -5,7 +5,7 @@ import { observer, useObservableEffect } from "../src"
 
 afterEach(cleanup)
 
-test("useObservableEffect", async () => {
+test("reactions run and dispose properly", async () => {
     let reactions1 = 0
     let reactions2 = 0
     let renders = 0
@@ -48,7 +48,7 @@ test("useObservableEffect", async () => {
         renders++
         return (
             <div>
-                `${props.store.prop1} ${props.store.prop2}`
+                {props.store.prop1} {props.store.prop2}
             </div>
         )
     })
@@ -78,4 +78,23 @@ test("useObservableEffect", async () => {
     expect(renders).toBe(3)
     expect(reactions1).toBe(1)
     expect(reactions2).toBe(1)
+})
+
+test("an undefined disposer works", async () => {
+    let renders = 0
+
+    const Component = observer(() => {
+        useObservableEffect(() => {
+            return undefined as any
+        })
+
+        renders++
+        return <div>test</div>
+    })
+
+    const { rerender } = render(<Component />)
+    expect(renders).toBe(1)
+
+    rerender(<div />)
+    expect(renders).toBe(1)
 })
--- a/test/useObservableEffect.test.tsx
+++ b/test/useObservableEffect.test.tsx
@@ -9,7 +9,7 @@ test("useObservableEffect", async () => {
     let reactions1 = 0
     let reactions2 = 0
     let renders = 0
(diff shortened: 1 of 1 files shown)


FredyC:
Nope, coverage still reports that condition has never fired. Sadly, I don't have enough permissions to enable Coveralls for this repo, so I can only run it locally.

```
-------------------------|----------|----------|----------|----------|-------------------|
File                     |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
-------------------------|----------|----------|----------|----------|-------------------|
All files                |    89.42 |    51.61 |    95.45 |    88.17 |                   |
 mobx-react              |      100 |      100 |      100 |      100 |                   |  
  jest.setup.js          |      100 |      100 |      100 |      100 |                   |
 mobx-react/src          |    89.32 |    51.61 |    95.45 |    88.04 |                   |
  ObserverComponent.ts   |    43.75 |    14.29 |       50 |    43.75 |... 35,36,41,42,44 |
  assertEnvironment.ts   |    66.67 |       50 |      100 |    66.67 |               5,8 |
  index.ts               |      100 |      100 |      100 |      100 |                   |
  observer.ts            |      100 |      100 |      100 |      100 |                   |
  staticRendering.ts     |      100 |      100 |      100 |      100 |                   |
  useComputed.ts         |      100 |      100 |      100 |      100 |                   |
  useObservable.ts       |      100 |      100 |      100 |      100 |                   |
  useObservableEffect.ts |      100 |    66.67 |      100 |      100 |                26 | <---
  useObserver.ts         |      100 |      100 |      100 |      100 |                   |
  utils.ts               |      100 |      100 |      100 |      100 |                   |
-------------------------|----------|----------|----------|----------|-------------------|
```

xaviergonz:
well, if the disposer doesn't get called I don't see how the variable with the number of calls to the disposer turns into two
Maybe coveralls + react hooks don't get along well?

FredyC:
> Maybe coveralls + react hooks don't get along well?

That's unlikely since all other files are just fine. I am not really sure what's wrong here.

xaviergonz:
well, I just put a console log in line 26 and it shows twice...

xaviergonz:
Using the code in this PR I mean

xaviergonz:
maybe what coveralls means is that the if condition is never false since the disposer is always available?

FredyC:
> maybe what coveralls means is that the if condition is never false since the disposer is always available?

Ah, that will be it :) So just for sake of coverage, there should be a test where it does not return a disposer and disable typescript check for that.

xaviergonz:
hopefully that last push solves that

FredyC:
Thanks, coverage is happy now :)

Btw, you have more permissions to enable https://coveralls.io/ for this repo?

xaviergonz:
Nope :( I can just ask for it, but someone else has to approve it

PR #34, opened 2018-12-04
First pass at creating a basic todo app as a complete example.

### LINKED ISSUE #33: Include runnable example in source code?
I've managed to successfully migrate a mobx v5  and mobx-react project over to mobx-react-lite, and was pleasantly surprisedly how relatively easy it was.

That being said, I can see the process being daunting for people that aren't as familiar with functional composition, use of context, etc.

@FredyC is there any desire to have a runnable example project? If so, I will happily volunteer to create something that can serve as way to complement the existing documentation in the readme.

larrymyers:
This still needs some polish and isn't ready for merge yet. Wanted to get an early PR created to get some feedback.

@FredyC is this something you think will be valuable to include directly in the project?

Closes #33

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,12 @@
 /node_modules/
 
+/example/node_modules/
+/example/yarn.lock
+
 dist/
 coverage/
 
+.idea/
 .rpt2_cache
 .DS_Store
 
--- a/example/package.json
+++ b/example/package.json
@@ -1,0 +1,12 @@
+{
+    "name": "mobx-react-lite-example",
+    "version": "0.0.1",
+    "main": "index.js",
+    "private": true,
+    "devDependencies": {
+        "ts-loader": "^5.3.1",
+        "webpack": "^4.26.1",
+        "webpack-cli": "^3.1.2",
+        "webpack-dev-server": "^3.1.10"
+    }
+}
--- a/example/src/todoForm.tsx
+++ b/example/src/todoForm.tsx
@@ -1,0 +1,33 @@
+import React, {FormEvent, useContext, useEffect} from 'react';
+import {observer} from '../../src';
+import {TodoStoreContext} from './store';
+
+export const TodoForm = observer(() => {
+    const todoStore = useContext(TodoStoreContext);
+    const input = React.createRef<HTMLInputElement>();
+
+    useEffect(() => {
+        if (input.current) {
+            input.current.focus();
+        }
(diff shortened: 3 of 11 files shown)


.gitignore
```diff
@@ -1,8 +1,12 @@
 /node_modules/
 
+/example/node_modules/
```
FredyC: I don't think you need these two, it should work recursively. Unless you have some IDE that would have a problem with it? In that case, I would rather create separate `.gitignore` in an `example` folder.

example/package.json
```diff
@@ -0,0 +1,12 @@
+{
+    "name": "mobx-react-lite-example",
+    "version": "0.0.1",
+    "main": "index.js",
```
FredyC: I don't think there should be a `main` here as it's not a module to be consumed.

example/src/todoForm.tsx
```diff
@@ -0,0 +1,33 @@
...
+    const todoStore = useContext(TodoStoreContext);
+    const input = React.createRef<HTMLInputElement>();
+
+    useEffect(() => {
+        if (input.current) {
+            input.current.focus();
+        }
+    });
+
+    const addTodo = (evt: FormEvent) => {
```
FredyC: What about `useCallback` here?

.gitignore
larrymyers: the leading leading slash only matches root of the project, so just the top level node_modules

example/package.json
larrymyers: yeah, I'll trim that out, just came with the "init" command from yarn.

example/src/todoForm.tsx
larrymyers: Good idea. I'm still working my way through all the new hooks included in React out of the box.

larrymyers:
@FredyC Thanks for the feedback. I'll go ahead and make more revisions to include more of the API in a way that makes sense so the example provides more value to new users of the project.

.gitignore
FredyC: Alright, missed that one, let's change it to `node_modules` then, I don't see a reason why not.

RayLuxembourg:
I have tried to implement this code, but change detection have not been triggered.

larrymyers:
@FredyC I'm going to close this PR. I'm not making progress on it and it looks like most people are looking for smaller, more concise examples anyway.

PR #35, opened 2018-12-06
Fixed typo in README

ross-weir:
Not sure if this tiny typo requires a PR :-)

Added missing bracket in code example for "What about smart/dumb components?" section of README


FredyC:
Thanks :) I am not even sure that example works, it was just quick & dirty attempt to show some possible approaches.

PR #37, opened 2018-12-09
Fix README inject example

### LINKED ISSUE #36: HOC inject example
Hi all.

I've been trying to find a way to create an `inject` helper function like in the README:

```
function inject(useSelector, baseComponent) {
    const store = useContext(StoreContext)
    const selected = useSelector(store)
    // optional memo essentially making a pure component
    return  baseComponent({ ...selected })
}
```

React throws a `Hooks can only be called inside the body of a function component.` error. I'm a bit of a noobie but haven't been able to find a way around this. It seems we may have to just use `useContext` and then a selector in the components themselves.

Can I please get your thoughts on this? Is my conclusion correct? Thanks :)

ross-weir:
fixes #36

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,7 +311,7 @@ const inject = (useSelector, baseComponent) => (
     React.useMemo((props) => {
         const store = useContext(StoreContext)
         const selected = useSelector(store)
-        // optional memo essentially making a pure component
+        
         return baseComponent({ ...selected, ...props })
     })
 )
--- a/README.md
+++ b/README.md
@@ -302,17 +302,19 @@ export default () => {
 }
 ```
 
-It may look a bit more verbose than a _classic_ inject, but there is nothing stopping you to make your own `inject` HOC which is so much easier since everything is just a funciton.
+It may look a bit more verbose than a _classic_ inject, but there is nothing stopping you to make your own `inject` HOC which is so much easier since everything is just a function.
 
 ```tsx
 // make universal HOC
 
-function inject(useSelector, baseComponent) {
-    const store = useContext(StoreContext)
-    const selected = useSelector(store)
-    // optional memo essentially making a pure component
-    return React.memo(props => baseComponent({ ...selected, ...props }))
-}
+const inject = (useSelector, baseComponent) => (
+    () => {
+        const store = useContext(StoreContext)
+        const selected = useSelector(store)
+        // optional memo essentially making a pure component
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -292,27 +292,31 @@ function UiComponent({ name, age }) {
...
-export default () => {
-    // you may extract these two lines into a custom hook
-    const store = useContext(StoreContext)
-    const data = userSelector(store)
-    return UiComponent({...data})
-    // perhaps wrap it inside observer in here?
-    return observer(UiComponent({...data}))
-}
+export default () => (
+    () => {
```
FredyC: Well, this piece is correct actually :) It's just a regular component with hooks, it shouldn't be returning a function.

README.md
```diff
@@ -292,27 +292,31 @@ function UiComponent({ name, age }) {
...
 // make universal HOC
 
-function inject(useSelector, baseComponent) {
-    const store = useContext(StoreContext)
-    const selected = useSelector(store)
-    // optional memo essentially making a pure component
-    return React.memo(props => baseComponent({ ...selected, ...props }))
-}
+const inject = (useSelector, baseComponent) => (
+    () => {
```
FredyC: I believe it should actually like this. With your approach, it's a component returning another component.

```suggestion
    React.useMemo((props) => {
        const store = useContext(StoreContext)
        const selected = useSelector(store)
        return baseComponent({ ...selected, ...props })
    })
```

ross-weir: Thank you. Hopefully it's looking OK now :+1:

FredyC: Copy & paste again, you forgot one thing. And try to think about what it actually does :)

ross-weir: Ops. Thanks for your patience. I'll be sure to do a lot more research!

README.md
```diff
@@ -302,17 +302,19 @@ export default () => {
...
-    const store = useContext(StoreContext)
-    const selected = useSelector(store)
-    // optional memo essentially making a pure component
-    return React.memo(props => baseComponent({ ...selected, ...props }))
-}
+const inject = (useSelector, baseComponent) => (
+    React.useMemo((props) => {
+        const store = useContext(StoreContext)
+        const selected = useSelector(store)
+        // optional memo essentially making a pure component
```
FredyC: And this comment doesn't really fit here, does it? :)

PR #43, opened 2019-01-09
1.create observable by function

ilovedesert001:


FredyC:
@ilovedesert001 You should have probably created an issue first. What is your reasoning behind adding this? I mean, for simple objects it's most likely unmeasurable optimization to wrap into a function. Cannot imagine you would be putting something computation heavy to initial state.

ilovedesert001:
Because I need something like class Component constructor

FredyC:
Can you elaborate more? I don't follow how is that relevant.

PR #49, opened 2019-01-25
[useObservable] optimize performance when per render.

alsotang:
avoid call `mobx.observable` every render. support lazy initialization


FredyC:
Can you explain why? This has been here recently in #43 already and I would really like to know why. Is this some kind of premature optimization?

mweststrate:
@FredyC, my guess is that the idea is borrowed from `useState`. The idea is that the initial value might be somethign expensive to construct (e.g. an object with 20 attributes, which are looked up all over the place), but never used be except for the initial render. However, it is created during every render, so just creating a lot of junk work for the GC + the initial construction costs. 

When passing a function, this work can be avoided. As the only work done on each render is creating that function + GC-ing it, which is at some point cheaper than construction those objects. Especially since that initialization function could be hoisted (in cases it doesn't depend on props / state), making it really efficient. So this is indeed a hook that makes additional optimizations possible.

src/useObservable.ts
```diff
@@ -3,6 +3,14 @@ import { useRef } from "react"
 
 type SupportedValues = object | Map<any, any> | any[]
 
-export function useObservable<T extends SupportedValues>(initialValue: T): T {
-    return useRef(observable(initialValue)).current
+export function useObservable<T extends SupportedValues>(initialValue: T | (() => T)): T {
+    const observableRef = useRef((null as unknown) as T)
```
FredyC: What about this way? Feels awkward double typing like that.
```suggestion
    const observableRef = useRef<T | null>(null)
```

test/useObservable.test.tsx
```diff
@@ -76,7 +76,7 @@ describe("is used to keep observable within component body", () => {
                 x: 1,
                 y: 2,
                 get z() {
-                    return obs.x + obs.y
+                    return this.x + this.y
```
FredyC: Please revert this, it's intentional like this. You should be able to access values without referencing `this`.

FredyC:
@alsotang Please update README as well.

@mweststrate Thanks for the explanation. I do understand what lazy initialization means, but it feels like an unnecessary complication to the API while it's not really measured if such optimization helps anything. I doubt that `useObservable` will be ever used for some huge state objects, there are better tools for that. Personally, I considered it more like an observable alternative to `useState`.

alsotang:
The explanation of @mweststrate is very correct. I express my thought below.

1. I think `anonymous function lazy initialization` is easy to understand, so ignore it. Just like `useState` api.
2. `useRef` only use the init value at the first component render. When the component is rendered, the init value is ignore. But we still compute the `mobx.observable(initValue)` again and again. It's unnecessary and waste.

So above two points are both about lazy initialization that save our compute resource.

alsotang:
I think this optimization would not add complexity to api, but only to api implement and can help some heavy usage scene.
The only problem now is the ts infer problem above. The `obs` case.

alsotang:
I think it's a ts weakness, not my fault. So now I can only implement the lazy `useRef` assign, but ignore the `anonymous function lazy initialization`

alsotang:
Now I change my pr to only implement lazy `useRef` assign. Remove code about `anonymous function lazy initialization`.

joenoon:
@alsotang I liked your original proposal with lazy evaluation similar to useState for the reasons @mweststrate explained.

I worked around this by actually using useState, but it would be great if useObservable could take a lazy function instead.  I don't have my code in front of me, but it looked something like this from memory:

```
// before
const {data} = useQuery(graphqlQuery); 
const store = useObservable(new Store(data)); // creating a new Store on every pass, expensive
```

```
// after
const {data} = useQuery(graphqlQuery); 
const [_store] = useState(() => new Store(data)); // only create a new Store once
const store = useObservable(_store); // cheap
```

With your original proposal, I think it would simplify to:

```
// before
const {data} = useQuery(graphqlQuery); 
const store = useObservable(() => new Store(data)); // only create a new Store once, cheap
```

If the TS typings are the only problem, maybe we can look at what useState does as that seems to work well for me with a lazy function resulting in the proper type.  I can look at this when I get back to my personal computer tonight as well.

alsotang:
@joenoon Great! Your work around by using `useState` inspired me. At least I can work around this `anonymous function lazy initialization` problem.

And I try many tricks of ts this all night to make ts work with it. But I failed.

FredyC:
@joenoon 
> // before const {data} = useQuery(graphqlQuery); const store = useObservable(new Store(data)); // creating a new Store on every pass, expensive

What is the `Store` actually? Isn't it an observable already? In that case, you should just `useState` to keep a reference to it. If it's just a plain class, I cannot imagine why you do that. If necessary to do some processing, factory function is always better imo.

@alsotang I think it's good middleground to do it like this. I still feel that lazy initialization is a premature optimization. And if someone wants to have a really expensive initial state there, it's always possible to have a custom hook since the `useObservable` is just a one-liner. I might even add the explanation to the README.

joenoon:
@FredyC I'm open to suggestions.  I've been constantly refactoring and trying different strategies to get to a point that feels cleanest/simplest.  I don't have my exact code in front of me, but it goes something like this (I didn't add all the typescript here, but ts is one of the main benefits):

```
class WorksheetItemsStore {
  @observable status = null;
  @observable data = null;
  constructor(data) {
    this.data = data;
  }
  get someDeepValue() { return this.data.foo.bar.x; }
  save = async () => {
    this.status = 'saving';
    await doSomething();
    this.status = null;
  };
  // etc...
}
```

The basic idea being to contain the logic specific to a certain input type (data).  So I pass down WorksheetItemsStore to components who need to operate on this "model".  Its a break away from the "one-store" way to be sure.  So far I find it to be more isolated/testable/type-able.

I like your point about being able to simply add a custom hook myself with my useState logic.  So if this isn't in the core its really easy to add and use myself.  I guess it just becomes a question then of how common it is.  Like I said I'm still trying various approaches to come up with the simplest pattern myself.

joenoon:
Yes I've looked at MST and will even give it another look.  I've also looked at the new useReducer hook, the Immer library, useImmer, redux etc.  In fact when I started looking into hooks I wondered if it would make mobx unnecessary and if I could get away with useState/useReducer alone (the answer is no, mobx is still necessary, and even less code with hooks). So far, I've found the above to be the simplest and clearest.  I've actually refactored one of my side project apps in each way to draw comparisons.  I've found this approach so far to be shorter than reducers/redux by at least 10%... on a bigger and more realistic application mobx would probably allow for a 20-30% code reduction.  And beyond lines of code, its also much more understandable - its simply an instance with expected type checking that components can interact with.  But I do want to take another look at MST - my concern is if its that redux-style of indirection thats really not what I'm looking for.  I know there is a purist angle to the whole state action new state which I get, and am fine with if it happens under the hood (ie. not in my app code), but I've found that style to be way too much noise for my taste.  I'm looking for ways of less code, simpler code.  That said if it were to reduce my code, be as or more simple, and have great type checking, then I'd jump onboard right away.

FredyC:
MST is certainly far from the Redux. It's just a nicer way of defining stores which are observable on its own (no need to wrap into `mobx.observable` and besides actions and computed props it has advanced features like snapshots, patching and things I don't even know about yet :) It might feel overwhelming at first, but it's worth it for sure.

Nonetheless, I am going to merge this as it surely provides some improvements even without measurement and doesn't break anything.

FredyC:
Btw, there is one realization that might not be apparent and wasn't possible with `mobx-react`. With `useObserver` the state doesn't even need to be recreated on every state change, because only a part of the component gets rerendered. I've updated a [README example](https://github.com/mobxjs/mobx-react-lite#useobservabletinitialvalue-t-t) that's showing that.

This change just got released under 0.3.6

FredyC:
I think this is a good place as any to talk about, but for future reference, it's probably better to open a new issue for a discussion.

I am not entirely sure what's the benefit of `useMemo` in this case. There has been an attempt in the past to have an observable updated when eg. props changes, but it fairly expensive to do so. Recreating a whole observable based on external a change sounds like a bad idea as it would throw away. Ultimately, it turned out better to have `useComputed` which can rely on external values.

PR #51, opened 2019-01-27
use-observer memory leaks

alsotang:
every time we call useObserver, there is a new `new Reaction` created, but it doesn't dispose

### DIFF
--- a/src/useDisposable.ts
+++ b/src/useDisposable.ts
@@ -20,7 +20,7 @@ export function useDisposable<D extends TDisposable>(
     disposerGenerator: () => D,
     inputs: ReadonlyArray<any> = []
 ): D {
-    const disposerRef = useRef<D | undefined>(undefined)
+    const disposerRef = useRef<D | null>(null)
     const earlyDisposedRef = useRef(false)
 
     useEffect(() => {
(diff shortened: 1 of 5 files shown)


FredyC:
Nice catch! I am wondering if the test case can be made for this. It's implementation detail, but it would be nice to cover it somehow.

I am please fix linter issue (see travis).

alsotang:
Yes, test should cover this. Which mock lib do you prefer? I always use this simple one: https://github.com/node-modules/mm

alsotang:
I will add some code.
Change all `useRef(undefined)` to `useRef(null)`, because react official docs use the latter.

FredyC:
> Yes, test should cover this. Which mock lib do you prefer? I always use this simple one: https://github.com/node-modules/mm

I am not sure how that module would help you write a test. Besides Jest already has a capability for [mocking modules](https://jestjs.io/docs/en/manual-mocks). Point is that test would somehow need to watch for creation and disposal of Reactions. It's pretty much implementation detail as I said. I am not sure on this one.

alsotang:
I would take look at jest framework. I dont familiar with it now.

alsotang:
I've write test cases for this pr and https://github.com/mobxjs/mobx-react-lite/pull/49

alsotang:
@FredyC any response?

FredyC:
LGTM. Thanks.

PR #54, opened 2019-02-01
Updating to React 16.8.0-alpha.1

dimitarnestorov:
`useImperativeMethods` was renamed to `useImperativeHandle`


alsotang:
LGTM. But lockfile is complex and vulnerable for hacker, so we should generate it ourselves.

FredyC:
I think I would rather rewrite that test. It's unnecesary to use that in there. I don't want to update those dependencies at all. I am sure there are still some projects (including mine) that are running early 16.7 alpha.

FredyC:
Closing in favor #57, thanks for contribution.

Issue #53, opened 2019-02-01
[On hold] React Concurrent mode: avoid side effects in render


SomeHats:
Hi! I've been really enjoying mobx-react-lite, thank you so much for this library. I've recently been doing some investigation about using MobX with suspense in react and have noticed some issues with the current useObserver implementation.

First of - components can and will throw. Suspense relies on throwing a promise. I'm a little unsure how MobX handles reactions throwing in general, but right now if you have an observed component and it throws, that upsets MobX and your thrown exception can't be handled in the usual way React does that - with error boundaries and suspense boundaries. That's a relatively simple issue to solve though.

The more serious issue is that once a component throws, React will often completely discard the tree up to the nearest boundary. That means all hooks get rolled back - so you can't rely on each render resulting in a subsequent commit (where useEffect hooks are called and their cleanup functions registered). In general, with the upcoming Concurrent mode, [render will often get called more than once](https://reactjs.org/docs/strict-mode.html#detecting-unexpected-side-effects).

This can be demonstrated fairly easily with a component like this:
```js
import React from "react";
import ReactDOM from "react-dom";

let isFirstRun = true;
function Suspend() {
  console.log("render");

  if (isFirstRun) {
    isFirstRun = false;
    console.log("suspend");
    throw new Promise(resolve => setTimeout(resolve(1000)));
  }

  React.useEffect(() => {
    console.log("mounted");
    return () => {
      console.log("unmounted");
    };
  }, []);

  return null;
}

function App() {
  return (
    <React.Suspense fallback={<div />}>
      <Suspend />
    </React.Suspense>
  );
}

const rootElement = document.getElementById("root");
ReactDOM.render(<App />, rootElement);
```
[CodeSandbox link](https://codesandbox.io/s/wqz6r87y87)

This component logs the following to the console:
```
render
suspend
render
mounted
```

So, what does this mean for mobx-react-lite? Right now, `useObserver` works by creating a reaction in the render phase, then cleaning up when the component unmounts:
https://github.com/mobxjs/mobx-react-lite/blob/d46cea3a7f812f75ec5582237806b084f0b70828/src/useObserver.ts#L24-L32

But each `render` call isn't guaranteed to come with a subsequent mount & render. The moment an observed component throws, all the effects of hooks (including the `useRef` for keeping track of the reaction) get discarded, so we end up leaking reactions all over the place.

[...]

PR #56, opened 2019-02-04
Rethrow catched exceptions

mweststrate:
See #53, exceptions catched by MobX's Reaction are now catched and rethrown, rather than just logged.

### DIFF
--- a/test/observer.test.tsx
+++ b/test/observer.test.tsx
@@ -466,7 +466,7 @@ function runTestSuite(mode: "observer" | "useObserver") {
                 act(() => {
                     x.set(42)
                 })
-                expect(errorsSeen).toEqual(["The meaning of life!"])
+                expect(errorsSeen).toEqual(["Error: The meaning of life!"])
                 expect(rendered.container.querySelector("span")!.innerHTML).toBe("Saw error!")
             } finally {
                 // tslint:disable-next-line
--- a/test/observer.test.tsx
+++ b/test/observer.test.tsx
@@ -1,3 +1,4 @@
+import mockConsole from "jest-mock-console"
 import * as mobx from "mobx"
 import * as React from "react"
 import { act, cleanup, fireEvent, render } from "react-testing-library"
@@ -450,12 +451,7 @@ function runTestSuite(mode: "observer" | "useObserver") {
                 return <span>{x.get()}</span>
             })
 
-            // tslint:disable-next-line
-            const origErrorLogger = console.error
-            // tslint:disable-next-line
-            console.error = function() {
-                // supress printing warnings that React always prints in DEV, even with error boundaries...
-            }
+            const restoreConsole = mockConsole()
             try {
                 const rendered = render(
                     <ErrorBoundary>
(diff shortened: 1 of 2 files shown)


mweststrate:
Sure! (sorry, forgot to mark as WIP)

Op ma 4 feb. 2019 om 16:28 schreef Daniel K. <notifications@github.com>:

mweststrate:
@FredyC rebased and added unit test

test/observer.test.tsx
```diff
@@ -416,6 +416,64 @@ function runTestSuite(mode: "observer" | "useObserver") {
...
+
+            const C = obsComponent(() => {
+                if (x.get() === 42) {
+                    throw new Error("The meaning of life!")
+                }
+                return <span>{x.get()}</span>
+            })
+
+            // tslint:disable-next-line
+            const origErrorLogger = console.error
```
FredyC: Please replace with mockConsole, used eg. in useDisposable test

FredyC: I am wondering though, I tried running it locally without suppressing and why it would be showing it in a console it was caught? I wasn't fiddling with error boundaries too much yet but feels strange.

test/observer.test.tsx
```diff
@@ -416,6 +416,64 @@ function runTestSuite(mode: "observer" | "useObserver") {
...
         })
     })
+
+    describe("error handling", () => {
+        test("errors should propagate", () => {
+            const x = mobx.observable.box(1)
+            const errorsSeen: any[] = []
+
+            class ErrorBoundary extends React.Component {
+                public static getDerivedStateFromError() {
```
FredyC: I was about to suggest to modify lint rules instead, but ok :)

mweststrate:
Pro tip: with this history better squash merge it ;-)

test/observer.test.tsx
```diff
@@ -416,6 +416,64 @@ function runTestSuite(mode: "observer" | "useObserver") {
...
+                const rendered = render(
+                    <ErrorBoundary>
+                        <C />
+                    </ErrorBoundary>
+                )
+                expect(rendered.container.querySelector("span")!.innerHTML).toBe("1")
+                act(() => {
+                    x.set(42)
+                })
+                expect(errorsSeen).toEqual(["Error: The meaning of life!"])
```
FredyC:
```suggestion
                expect(errorsSeen).toMatchSnapshot(mode)
```

test/observer.test.tsx
mweststrate: It *is* really weird indeed, see also: https://github.com/facebook/react/issues/10474

The fun thing, the error doesn't escape any more (without the `<ErrorBoundary>` component you would correctly get an exception on `x.set(42)`), yet, it is _still_ being logged. 

I think it is meant as additional help to DEVs to see errors early. (After all error boundaries are for _unexpected_ exceptions, not normal errors). But in case of a test runner it is pretty annoying behavior.

test/observer.test.tsx
mweststrate: meh, it makes the test itself harder to read I think? (I usually favor `toMatchInlineSnapshot` over `toMatchSnapshot` for small stuff, but that wouldn't help much in this case)

mweststrate: But if you prefer I'll replace it

test/observer.test.tsx
mweststrate: Updated with `mockConsole`. Neat.

test/observer.test.tsx
FredyC: I tried inline snapshot, but in this case it doesn't work because the test suite is executed twice and it doesn't like it :)

PR #57, opened 2019-02-06
[WIP] Try update to React 16.8

### LINKED ISSUE #58: mobx-react-lite 0.3.7 does not work with react 16.8.0
I tested out an upgrade on react app I'm building from 16.8.0-alpha.1 to 16.8.0.   Best I can tell on some initial analysis, it looks like function components wrapped in observer() aren't rendering when observables are modified in an action.

I verified that this issue is present all the way back to 0.3.4, so it's highly likely that something in the linked changelog below has resulted in a breaking change with mobx-react-lite.

https://reactjs.org/blog/2019/02/06/react-v16.8.0.html#hooks-changelog-since-alpha-versions

I went ahead and created a branch on my mobx-react-lite fork to test the 16.8.0 upgrade, and the result is most of the tests run by jest fail as well.

I'll spend some time digging into what the breaking change is and post any updates to this issue.

FredyC:
So [16.8 is officially out](https://reactjs.org/blog/2019/02/06/react-v16.8.0.html) with hooks. I tried updating it and a bunch of tests got broken. I will look into it later or feel free to help with that.

Once it's fixed, I want to release this as 1.0.0. There won't be probably any API change for now anyway.

Fixes #58


FredyC:
Ok, so fixing the tests was fairly easy after all. The `useState` apparently got optimized and it's no longer rerendering if a value hasn't actually changed. I had to revert our `forceUpdate` utility function which modification kinda slipped away in recent changes by @alsotang. 

_Wondering if there is actually some hooks variant of `forceUpdate`, haven't seen any mention of it._

No we basically need to wait for react-testing-library to [merge use of `act` utility](https://github.com/kentcdodds/react-testing-library/pull/278) to get rid of those warnings spilled all over the tests.

FredyC:
I've also upgraded other dev dependencies so we are coming to 1.0.0 with a clean slate. Should really setup some dependencies auto update app - recommendations?

Jest will stay at 23 for now because `ts-jest` support is not there yet.

xaviergonz:
> Should really setup some dependencies auto update app - recommendations?

`yarn upgrade-interactive` ?

src/utils.ts
```diff
@@ -7,10 +7,10 @@ export function useUnmount(fn: () => void) {
 }
 
 export function useForceUpdate() {
-    const [, setTick] = useState(null)
+    const [, setTick] = useState(0)
 
     const update = useCallback(() => {
-        setTick(null)
+        setTick(tick => tick + 1)
```
xaviergonz: I know this is most likely not an issue at all, but we "only" have up to 9007199254740991 (Number.MAX_SAFE_INTEGER) renders until this breaks :)
Might be worth to set 0 once that value is reached?

test/useDisposable.test.tsx
```diff
@@ -106,7 +106,6 @@ test("reactions run and dispose properly", async () => {
     expect(reaction2DisposerCalls).toBe(0) // this one is not early disposed
 
     rerender(<Component store={store} a={1} />)
-    flushEffects()
```
xaviergonz: they finally fixed that on react-testing-library? nice!

xaviergonz:
Btw, do the error console logs actually mean we should be using act on unit tests or are the react-testing-library guys going to fix that somehow?

src/utils.ts
FredyC: Um, would really like to see an app that would have a single component rendered that long to waste that :D I am mostly hoping that some official way of `forceUpdate` will rise eventually.

test/useDisposable.test.tsx
FredyC: It's more of the fix in React itself :) https://reactjs.org/docs/test-utils.html#act

xaviergonz:
Nevermind, just saw the comment about react-testing-library

FredyC:
@xaviergonz That's what I am trying to figure in here: https://github.com/kentcdodds/react-testing-library/issues/281

> > Should really setup some dependencies auto update app - recommendations?
> 
> `yarn upgrade-interactive` ?

I mean something automatic. I do recall some Greenkeeper tool from the past, but not sure if there is something better.

src/utils.ts
xaviergonz: Makes me wonder if flipping a boolean would be enough or if react is smart enough so if the bacthed state is not changed it won't re-render

xaviergonz: Either way, given a component is rendering 60 times per second it will take 285616414 years more or less for it to break, so it was just a stupid comment :)

FredyC:
Alright, tests are finally fixed to the top notch again. And React 16.8.1 got released meanwhile so it's there also :)

stnwk:
@FredyC 

> I mean something automatic. I do recall some Greenkeeper tool from the past, but not sure if there is something better.

Take a look at https://renovatebot.com/ - it's free for Open Source Projects :)

Also: Thanks for being so quick about fixing this! ❤️

Issue #65, opened 2019-02-12
Can't work with old webpack uglify plugin?


xwillmadeit:
```js
ERROR in Encountered an error while minifying js/vendor.891a096d.js:
SyntaxError: Unexpected token: keyword (const)
  27927 |
  27928 | function useObservable(initialValue) {
> 27929 |     const observableRef = Object(__WEBPACK_IMPORTED_MODULE_1_react__["useRef"])(null);
        |    ^
  27930 |     if (!observableRef.current) {
  27931 |         observableRef.current = Object(__WEBPACK_IMPORTED_MODULE_0_mobx__["observable"])(initialValue);
  27932 |     }
```

The build version of this package has es6 syntax，but i use webpack-parallel-uglify-plugin，so it breaks my app. Do I have to upgrade my uglify plugin or config my babel-loader to transpile it first?

FredyC:
Is probably related to #63, but yea, would be better if you can update your stack to support ES6. It's kinda strange that you would go with a bleeding edge like React Hooks, but would be still requiring ES5 😕 Alternatively you can try a new and maintained [minifier Terser](https://github.com/terser-js/terser). I am not sure if it has parallel webpack plugin though.

xwillmadeit:
> Is probably related to #63, but yea, would be better if you can update your stack to support ES6. It's kinda strange that you would go with a bleeding edge like React Hooks, but would be still requiring ES5 😕 Alternatively you can try a new and maintained [minifier Terser](https://github.com/terser-js/terser). I am not sure if it has parallel webpack plugin though.

I didn't think too much because react hook is fully backwards-compatible. I use webpack 3+ and can't use [terser-webpack-plugin](https://github.com/webpack-contrib/terser-webpack-plugin)，so its kinda embarrassed... I guess it's time to upgrade webpack to 4+

FredyC:
Yea, this package is meant for like a future proof stacks. Please do try an update and if it turns out impossible, you would need to wait (or contribute) for #63 to get fixed. I will close this now and subscribe there, please.

alsotang:
uglify can not work with es6. It only support es5. First you should compile the module to es5, then uglify it.

PR #66, opened 2019-02-14
Changed build target to ES5, fixes #63, #65

### LINKED ISSUE #63: Dist won't work in IE11 due to "target": "es6" in tsconfig.json
Can tsconfig.json target es5 instead of es6? Currently when including mobx-react-lite in a CRA app the dist is built against es6 and will not run in IE11

mweststrate:
See #63 for details

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -2,7 +2,7 @@
     "version": "2.0.10",
     "compilerOptions": {
         "outDir": "dist",
-        "target": "es6",
+        "target": "es5",
         "module": "esnext",
         "moduleResolution": "node",
         "downlevelIteration": true,


stnwk:
Thank you very much for this :)

mweststrate:
@FredyC afaik there is currently only _one_ bundler that supports ES6 modules, metro. All other bundlers expect either the code to be delivered in ES5 (webpack, parcel, rollup, etc), or will do so themselves by default (for example CRA will transpile the ES6 constructions anyway, even in dev mode, if you inspect the sources). So es6 won't be preserved anyway.

The setup is already the same now as in MobX, React itself, etc. (React itself is shipped as es5 as well). The difference is really minor, so I don't see really the point of deliberately breaking the build setup of people, without clear benefit. 

Note that there is a difference between ES6 _syntax_ and ES6 _language library_. This MR only compiles the _syntax_ part back. (Strictly speaking; ES5 + ESM, as ESM is preserved)

FredyC:
It might surprise you (maybe not), but it's not about bundler what output you want to have. It's about loaders/plugins that are parsing that code and producing something out of it. 

In our company products, we do have it set up to produce ES6 code because we are deliberately ignoring IE11 (🖕) and the other major browsers can run ES6 code just fine.

Since it's not possible to transform ES5 to ES6, I would like to find a way how to keep ES6 and figure out why other people have issues. As you said, bundlers would produce ES5 code anyway. If it's problem of language, this PR won't help them for sure.

stnwk:
> In our company products, we do have it set up to produce ES6 code because we are deliberately ignoring IE11 (🖕) and the other major browsers can run ES6 code just fine.

Good for you 👌🏻

We are using `mobx` and `mobx-react-lite` to build our web app. Unfortunately, we still have some big customers who depend on IE11 support. It's not always fun and games and you're not always in the position to just say "fuck IE11". Many products and markets and especially companies with existing customers kind of **have to** support it, while still trying to push forward.

But by assuming everyone is in the same privileged position to just not care about it, you're really making things hard for people out there. Things don't just work™️ any more.

In the current way you are blocking us from using mobx and hooks together or forcing us to compile our node_modules. Neither of which I'm really happy about..

I really wished I wouldn't have had to write this comment. Please figure this out rather sooner than later.
Thank you!

FredyC:
@stnwk That wasn't a really helpful comment, you need to be more concrete what issues you actually have. As I just said in https://github.com/mobxjs/mobx-react-lite/issues/63#issuecomment-463609002 if your dev stack is setup do produce ES5, then this change won't do anything. Instead, you need to ensure you have proper polyfills for runtime code that IE11 is missing.

FredyC:
There might be a perhaps different set of problems that bundler/loader/plugin is not able to parse ES6, but that's an issue that would show during a build step. So far everyone is talking about code not working in IE11 which I am fairly certain is only a problem of polyfills.

If someone has actual build issues, it's perhaps time to look into updating the tooling? It's 2019, and ES5 is like really old now and we should be trying to move from there instead stagnating in the past.

mweststrate:
@FredyC I think the point you are missing, is that bundlers normally _don't_ transpile node_modules! Babel settings etc all apply to the source code of the project, not the source code of it's dependencies. So for your own code you can use whatever your want, but `node_modules` aren't transpiled, only _bundle_ (as the name suggests), for two reasons:

1. It would make every build significantly more expensive
2. Worse: there is no standard to configure what the correct transpilation settings are. For example, is an  array destructing written, assuming the the transpiler should use the iterator protocol or not? For example: `downLevelIteration` is a setting in _our_ tsconfig is a setting that would need to be picked up by the bundler to make sure that the library is compiled in the correct way to ES5. But that whole tsconfig isn't (or shouldn't) even be part of the NPM module. 

Here is a fun example: `mobx-react` won't even build with the current `mobx-react-lite` as dependency, because rollup simply doesn't transpile anything in the node_modules (for above reasons) and bails out on the first `let`. So even for ourselves we have to fix this, as otherwise we can't package mobx-react :smile: 

You could also inverse the question: what is the benefit of _not_ shipping ES5. As a consumer you shouldn't care that there are things in the module would read nicer if it was ES6. Did you ever run into any trouble because React ships ES5? The only real benefit I see is that the module woudld be marginally smaller. 

If you find it really important I propose the following; add an additional es6 syntax build target, which is not referred to by any entry point in `package.json`, and then if you deliberately want to use the ES6 version, you can set it up an alias in your build process `"mobx-react-lite": "mobx-react-lite/es6"` to have the neater version. This way it works out of the box for everybody, and people can opt-in for an optimized experience.

stnwk:
The error we are getting in IE11 is:
```
SCRIPT1006: Expected ')'
vendors.9b1a778c.chunk.js (88,68847) in index.module.js:108

// which is
function useObserver(fn, baseComponentName = "observed") {
```

The compile target is `es6` with `babel-preset-env` correctly setup, everything works in IE11 and is correctly compiled. However, since we do not transpile our `node_modules` we are shipping the `es6 dist` from `mobx-react-lite` to our customers, which breaks.

FredyC:
Ok, I get it now, I have indeed forgotten about that part that transpiling node_modules would be rather costly.

It's really fucked up situation ☹️ Just because there is ages old browser that Microsoft decided not to develop anymore and some BFU are still using it, we have to be shooting ourselves in the foot and ignore natively implemented ES6 features. Sure I admit that in case of this package it's a minimal problem, there isn't that much of the code, but it's ... sad.

> Here is a fun example: `mobx-react` won't even build with the current `mobx-react-lite` as dependency, because rollup simply doesn't transpile anything in the node_modules (for above reasons) and bails out on the first `let`. So even for ourselves we have to fix this, as otherwise we can't package mobx-react 😄

That's kinda weird because Rollup should consider that package as an external dependency (or you need to tell it) and it doesn't need to parse it all in that case. I am doing that all the time, bundling dependencies is bad because when you have that package installed separately, you are getting twice the same code.

Ok, let's merge it and ship it...

mweststrate:
> That's kinda weird because Rollup should consider that package as an external dependency (or you need to tell it) and it doesn't need to parse it all in that case. I am doing that all the time, bundling dependencies is bad because when you have that package installed separately, you are getting twice the same code.

@FredyC so far, definitely not set in stone, I was assumping that mobx-react would be mobx-react-lite + more, re-exporting all things that ship in `-lite`, so that people would always choose for one of those, and preventing annoying  mistakes like `import { observer  ]from "mobx-react-lite"; @observer class`...

alsotang:
ah...I also forget that people don't transplie node_modules by default setting.
So if we ship es6 module, then the language problem can not be realized in compile stage.
This pr seems good.

Issue #32, opened 2018-12-01
Question: hoist-non-react-statics


bringking:
First thanks for this library, I love the updates compared to the old package. When implementing this in a Next.js codebase, I ran into issues where the `observer` HOC doesn't hoist non-react-statics e.g. getInitialProps. Is this something we should add? If so, I would be happy to make a PR. 

If not, the workaround is easy enough 

```javascript
const ObservedComponent = observer(Component);
ObservedComponent.getInitialProps = () => { };
```

bringking:
@xaviergonz yeah, true `useObserver` might be just as good..

FredyC:
I don't know. Personally, I am no fan of using static properties like this. What's the point? Why it cannot be just a scoped function which you can export along the component if needed? Feels much cleaner in my opinion.

```js
const ObservedComponent = observer(Component);
export const getInitialProps = () => { };
export default ObservedComponent
```

In case you really need it for some weird use case scenario, the `useObserver` could be a better candidate for sure. However, we should probably add some dev-only warning that checks for non-react statics and recommends to `useObserver`. Can you prepare such PR @bringking ?

mweststrate:
Thinking about this, I think we should support hoisting for a few reasons:
1. Some frameworks depend on it, .e.g Next. It can be worked around, but I think finding out that something has not been hoisted is a subtle thing to find in the first thing (in this specific case: a lifecycle is not fired)
2. mobx-react's `observer` currently does supports hoisting for for function components, changing that would be quite a subtle breaking change

So I think there are two ways to fix it:
1. Only do hoisting in `mobx-react`'s observer, making it consistent with `observer` for classes and previous mobx-react versions, but inconsistent with `mobx-react-lite`
2. Do the hoisting in `mobx-react-lite` as a good HoC citizen here, and making the bundle slightly larger by it

FredyC:
Ok, I don't have any experience with Next and how are statics used there. If it's really needed, then I guess we have to obey 👼 

That module feels kinda bulky considering there are just three React related statics for a functional component. It's too much universal imo.

mweststrate:
> That module feels kinda bulky considering there are just three React related statics for a functional component. It's too much universal imo.

[...]

PR #70, opened 2019-02-22
Some basic features that will allow integration with mobx-react 6

mweststrate:


src/observer.ts
```diff
@@ -52,6 +57,22 @@ export function observer<P extends object, TRef = {}>(
         memoComponent = memo(wrappedComponent)
     }
 
-    memoComponent.displayName = baseComponentName
+    memoComponent.displayName = `observer(${baseComponentName})`
```
mweststrate: I'm not entirely sure about this change; it seems kinda consistent, since `observer` is basically a hook, (it was already a feature request on mobx-react 5 as well). On the other hand, one could also see from the hooks in the inspector that something is an observer. And it would liter `observer(XXX)` all over the place

src/observer.ts
```diff
@@ -52,6 +57,22 @@ export function observer<P extends object, TRef = {}>(
...
+    copyStaticProperties(baseComponent as any, memoComponent as any)
+
     return memoComponent
 }
+
+function copyStaticProperties(base: React.FunctionComponent, target: React.FunctionComponent) {
+    // From all the potential static properties in React, propTypes and defaultProps are the only ones that work
+    // (modern) for function components
+    // One could wonder whether custom defined static fields should be copied over.
+    // However, at this moment it doesn't seem common to establish custom fields on function components.
```
mweststrate: otherwise, we should bring in `react-hoist-non-statics` package, to make sure the right things get moved. But if feels kinda overkill, since the following isn't a pattern afaik:

```javascript
function Select() {

}

Select.Option = function() { // Some other component for example

}

export default observer(Select)
```

src/observer.ts
```diff
@@ -6,16 +6,21 @@ export interface IObserverOptions {
     readonly forwardRef?: boolean
 }
 
+export interface IMobXReactObserver {
+    isMobXReactObserver: true
```
FredyC: From what I gather it's just a hack for testing purposes? Do we really need to pollute each component with it?

src/observer.ts
FredyC: Yea, I think we don't really need it. As you say, it will be visible in hooks inspector and it's not like people need to know if a component is observed all the time. I would remove it.

xaviergonz: If it has to be there I'd rather see it as `Component (observer)`

src/observer.ts
```diff
@@ -52,6 +57,22 @@ export function observer<P extends object, TRef = {}>(
         memoComponent = memo(wrappedComponent)
     }
 
-    memoComponent.displayName = baseComponentName
+    memoComponent.displayName = `observer(${baseComponentName})`
+    copyStaticProperties(baseComponent as any, memoComponent as any)
```
FredyC: If I am not mistaken, we are manually setting `displayName` above and then again this function? Not a huge deal, but I would probably drop this method and just set the `propTypes` here directly. The comment feels unnecessary as well. Let's just refer to issue #32 in here instead of such essays :D

src/useObserver.ts
```diff
@@ -34,7 +35,15 @@ export function useObserver<T>(fn: () => T, baseComponentName = "observed"): T {
         }
     })
     if (exception) {
+        reaction.current.dispose()
```
FredyC: Isn't it already disposed on unmount? I am not sure if throwing counts as unmount, but it should.

test/observer.test.tsx
```diff
@@ -564,6 +564,20 @@ it("should only called new Reaction once", () => {
     spy.mockRestore()
 })
 
+it("should support hoisting", () => {
```
FredyC:
```suggestion
it("should hoist known statics only", () => {
```

FredyC: Also, add `propTypes` to the test to make the coverage happier :)

src/useObserver.ts
```diff
@@ -34,7 +35,15 @@ export function useObserver<T>(fn: () => T, baseComponentName = "observed"): T {
...
         throw exception // re-throw any exceptions catched during rendering
     }
     return rendering
 }
+
+function printDebugValue(v: React.MutableRefObject<Reaction | null>) {
```
FredyC: This is causing an annoying drop in the coverage. I would suggest separating it extra file and have some simple test for it. I doubt it can be tested from the component.

src/observer.ts
mweststrate: Agreed, I'll drop it

src/observer.ts
mweststrate: No, just to make external reflection by others possible. But without devtools this becomes less relevant indeed. I'm fine to drop it until, at least until somebody needs it again :).

src/useObserver.ts
mweststrate: Unmount is not triggered if it happened during the first render; as unmount is setup by `useEffect`, so it is not guaranteed to run if the component never mounted. However, the tracking (and hence establishing of observers) does happen in the first render.

mweststrate: (I realized this after the whole concurrent mode discussion)

src/useObserver.ts
mweststrate: Will do

src/useObserver.ts
FredyC: I assume the `Reaction` is fine with double calling `dispose` (if that happens)?

src/observer.ts
mweststrate: It doesn't seem to be overkill, see #32

src/observer.ts
mweststrate: Updated

src/observer.ts
mweststrate: Dropped

src/observer.ts
mweststrate: Dropped

src/useObserver.ts
mweststrate: Double dispose is no issue btw

src/useObserver.ts
mweststrate: Done

test/observer.test.tsx
mweststrate: Done

src/observer.ts
```diff
@@ -52,6 +52,19 @@ export function observer<P extends object, TRef = {}>(
         memoComponent = memo(wrappedComponent)
     }
 
+    copyStaticProperties(baseComponent as any, memoComponent as any)
```
FredyC: Why `any`? Wouldn't be easier to just inline that code instead of having a function that's not really reused? Or even inline a whole function if you want to name that operation, but you could use a closure that way.

mweststrate: The goal of the function is not reuse, it is to make the narrative clearer for future readers. `observer`'s goal is to wrap the function we're passing. As an essential detail we must hoist statics, but how that is done is really an implementation detail. Keeping functions small makes it much easier for future readers to grab what is globally happening, what is happening in detail, and what statements go together and provide some cohesive functionality.

FredyC:
Released 1.1.0

PR #96, opened 2019-03-23
shouldForceUpdate

xaviergonz:
Summary: a low level escape hatch for unwanted foruceUpdate calls

While making my own version of useObservableProps I thought of a way to skip unneeded double re-renders, which basically they happen because of this:
```
function C(props) {
  const obsProps = useObservableProps(props); // 1

  return useObserver(() => {
    <div>{obsProps.x}</div> // or a computed based on obsProps (2)
  })
}
```

first render:
- (1) no reaction yet established, so the change makes no reaction
- (2) obsProps (or a computed based on it) is marked as a dependency

second render:
- (1) it changes in order to update, which calls forceUpdate, which queues a future update
- (2) it gets rendered as usual
- sadly because of the forceUpdate from (1) it will render once more, even though nothing has actually changed between (1) and (2)

This PR includes a low level primitive (that might be hidden, but needs to be exported) to skip forceUpdate from doing anything in case it triggers in certain blocks, e.g.

```
function useObservableProps(props) {
  useSkipForceUpdate(() => {
    // here update observable props to props
    // this will actually trigger the forceUpdate reaction, but forceUpdate itself will be noop
  })
  // after it is done forceUpdate calls will work once more
}
```

it uses a global var for enable/disable (since there's no freaking way for two hooks to share state or get a component instance id (thanks react guys!)), but that should be good enough since it will run synchronously and is a rarely used low level optimization


src/utils.ts
```diff
@@ -6,12 +6,25 @@ export function useUnmount(fn: () => void) {
     useEffect(() => fn, EMPTY_ARRAY)
 }
 
+let skippingForceUpdate = 0
```
FredyC: Ugh, that doesn't sound right to be relying on a global variable...For multi-threading environments, it could cause some problems.

xaviergonz: Luckily then Js is not multithreaded really :)

xaviergonz:
I tried to think of many solutions...

- adding a second parameter to useObserver with a blacklist of variables it should not autoUpdate
  - could not get the list of variables that triggered a reaction (mobx doesn't offer it afaik) + it is kinda annoying to add obsProps to that array al the time

- use untracked anytime obsProps or derived values are used inside useObserver
  - would work, but kind of annoying and prone to breakage

- if react was able to get the component id and/or share states between hooks...
  - that'd be perfect, but as you know the react developers are totally against this (for whatever reason I don't get)

Therefore this (ugly) solution, but that should be safe since JS is not multi-threaded

That being said, I'm **really** open to alternative ideas.

FredyC:
Actually, looking at added tests, I don't follow what is the use case behind this. You are changing a state value in a render phase. That doesn't usually happen. I could imagine changing state based on user interaction, some timer, and similar activities. Doing that with regular `useState`, you would be causing an endless loop.

src/utils.ts
FredyC: Well, I heard that NextJS somehow is :)

xaviergonz:
I can change the tests if you want so it does the actual use case (updating an observable object from the props)

FredyC:
To be honest I am still not sure why you need it. React will re-render a component when props change and you cannot prevent that (unless you don't care about that update). Why would you need MobX to do another rendering on that change?

xaviergonz:
Check the updated tests, it should be more enlightening :)

xaviergonz:
The issue is that on second and subsequent renderings, without useSkipForceUpdate they will always render twice needlessly

xaviergonz:
Ah, another solution I thought was to add to observable objects a "flag" (symbol in the object or something) to flag them as "changes made to this object should skip forceUpdate", but for that to work I'd also need to know all the objects that triggered the reaction

@mweststrate is there a way using a Reaction class to know the objects/changes that triggered it?

FredyC:
> The issue is that on second and subsequent renderings, without useSkipForceUpdate they will always render twice needlessly

That's my point. Why don't you let React to handle rendering on props change? Why you want to employ MobX there and yet skip the re-render from it. I am either tired and seeing that usefulness in that ... or there is none :)

> Check the updated tests, it should be more enlightening :)

How is that test different? The state change is still happening within render phase, it's just wrapped inside the hook. I think it should happen within `useEffect` really ... it's a side effect after all.

xaviergonz:
To make a library that makes this https://github.com/mobxjs/mobx-react-lite/issues/94#issuecomment-475773636 work :)

In other words, to be able to use computeds, reactions, etc using observable props (the mobx way) rather than having to add props to the annoying react array of dependencies (which would end up re-creating reactions etc)

Actually it can be made already, but with the double renderings, so I wanted to figure a way to avoid those (and therefore here I am)

FredyC:
Hm, I am not really sold we should support that in here :) Feels super awkward to me.

However, the clean way to allow it could be an option for `useObserver` that accepts a custom `forceUpdate`. Then do your skip in there if I am not mistaken.

xaviergonz:
Hmm, that should work (as long as observer accepts a custom forceUpdate too to pass it to the internal useObserver)

Though I guess I'd need to provide a custom useObserver as well so people don't need to pass the custom forceUpdate as well, and a custom observer that auto passes that option... but better than nothing :)

xaviergonz:
@FredyC changed the PR based on your comment

src/observer.ts
```diff
@@ -1,9 +1,10 @@
 import { forwardRef, memo } from "react"
 import { isUsingStaticRendering } from "./staticRendering"
-import { useObserver } from "./useObserver"
+import { ForceUpdateHook, useObserver } from "./useObserver"
 
 export interface IObserverOptions {
     readonly forwardRef?: boolean
+    readonly customForceUpdateHook?: ForceUpdateHook
```
FredyC: Let's not pollute the `observer` with this. It's a low-level API and it should be on `useObserver` only. You can make your own `observer` for those needs.

src/useObserver.ts
```diff
@@ -4,15 +4,31 @@ import { printDebugValue } from "./printDebugValue"
 import { isUsingStaticRendering } from "./staticRendering"
 import { useForceUpdate, useUnmount } from "./utils"
 
-export function useObserver<T>(fn: () => T, baseComponentName = "observed"): T {
+export type ForceUpdateHook = () => () => void
+
+export function useObserver<T>(
+    fn: () => T,
+    baseComponentName?: string,
+    customForceUpdateHook?: ForceUpdateHook
```
FredyC: It's probably better to wrap this into a `options` object in case some other stuff comes later so it won't be a breaking change. Keep the `baseComponentName` as is.

src/useObserver.ts
```diff
@@ -4,15 +4,31 @@ import { printDebugValue } from "./printDebugValue"
...
         return fn()
     }
-    const forceUpdate = useForceUpdate()
+
+    const wantedForceUpdateHook = customForceUpdateHook || useForceUpdate
+    if (process.env.NODE_ENV !== "production") {
+        const forceUpdateRef = useRef<ForceUpdateHook>(wantedForceUpdateHook)
+        if (forceUpdateRef.current !== wantedForceUpdateHook) {
+            throw new Error(
+                "a custom force update hook cannot be switched to another one once used"
```
FredyC: I don't like this. It's an extra check for everyone just because there is a potential for a custom update hook. Is this an actual problem? You haven't captured it in the tests. I would say it's more about the rules of hooks. If someone would be that silly and change called hooks conditionally, they should burn in their own hell :)

src/useObserver.ts
```diff
@@ -4,15 +4,31 @@ import { printDebugValue } from "./printDebugValue"
 import { isUsingStaticRendering } from "./staticRendering"
 import { useForceUpdate, useUnmount } from "./utils"
 
-export function useObserver<T>(fn: () => T, baseComponentName = "observed"): T {
+export type ForceUpdateHook = () => () => void
+
+export function useObserver<T>(
+    fn: () => T,
+    baseComponentName?: string,
```
FredyC: Please, the default name should stay here so it's more apparent. Do not hide down in the code.

test/customForceUpdate.test.tsx
```diff
@@ -0,0 +1,131 @@
+import mockConsole from "jest-mock-console"
+import { observable, runInAction } from "mobx"
+import * as React from "react"
+import { useCallback, useState } from "react"
+import { cleanup, render } from "react-testing-library"
+import { useForceUpdate, useObserver } from "../src"
+
+afterEach(cleanup)
+
+describe("is used to make calls to force update skip re-renderings on demand", () => {
```
FredyC: Some people are getting inspiration from tests on how to use the lib and I would consider these examples a bad practice. Setting state in a render phase just doesn't feel right. Feel free to get wild in your lib, but not here, please.

Let's have a simple test for passing that custom force update and validating it works.

src/useObserver.ts
FredyC: And actually same for that added option. It should be in here and down in the code. And you can rename that option to `useForceUpdate` directly, no need to call it a custom.

src/observer.ts
xaviergonz: The problem with that is basically that to make a custom observer would mean to copy and paste the whole thing anew :-/

FredyC: Yes, that's exactly that :) If you want to make crazy stuff you have to do something extra ;) Honestly, it's not such a big thing to copy.

test/customForceUpdate.test.tsx
xaviergonz: done, simplified

src/useObserver.ts
xaviergonz: done

src/useObserver.ts
xaviergonz: agreed, removed

src/useObserver.ts
xaviergonz: wrapped

src/observer.ts
xaviergonz: no, but then that'd mean that I'd need recopy and paste everytime observer changes and change one line :-/

src/useObserver.ts
FredyC: Thanks, much better.

src/observer.ts
FredyC: Sure, but on another hand, it's about the simplicity of the API. It's enough to allow this craziness in a low-level API, I don't like to spread more like that. Besides, I don't think that `observer` will be changing too much in the future, there is very little logic. The typings are the bulkiest part in here.

README.md
```diff
@@ -127,7 +127,7 @@ The following optional parameters are available:
 
 As for the options, the following are available:
 
--   `customForceUpdateHook`: pass a custom hook that should make a component re-render (or not) when changes are detected.
+-   `useForceUpdate`: pass a custom hook that should make a component re-render (or not) when changes are detected.
```
FredyC: Some mention about default should be here. It sounds like you need to specify this every time :)

src/useObserver.ts
```diff
@@ -7,24 +7,24 @@ import { useForceUpdate, useUnmount } from "./utils"
...
 export interface IUseObserverOptions {
-    customForceUpdateHook?: ForceUpdateHook
+    useForceUpdate?: ForceUpdateHook
 }
 
 export function useObserver<T>(
     fn: () => T,
-    baseComponentName?: string,
+    baseComponentName: string = "observed",
     options?: IUseObserverOptions
```
FredyC: If you do it like this, the default will be applied from the import and it's still possible to override.
```ts
options: IUseObserverOptions = {
  useForceUpdate
}
```

src/useObserver.ts
```diff
@@ -7,24 +7,24 @@ import { useForceUpdate, useUnmount } from "./utils"
...
 ): T {
     if (isUsingStaticRendering()) {
         return fn()
     }
 
-    const wantedForceUpdateHook = (options && options.customForceUpdateHook) || useForceUpdate
+    const wantedForceUpdateHook = (options && options.useForceUpdate) || useForceUpdate
     const forceUpdate = wantedForceUpdateHook()
```
FredyC: With default you can do this in here and nothing else.
```ts
const forceUpdate = options.useForceUpdate()
```

README.md
xaviergonz: done

src/useObserver.ts
xaviergonz: not quite, if that's done and a new property is ever added to options and options does not contain useForceUpdate then that would break

e..g
```
useObserver(..., {
  someOtherOption: ...,
  // no useForceUpdate since it is optional
}

const forceUpdate = options.useForceUpdate() // useForceUpdate will be undefined 
```

src/useObserver.ts
xaviergonz: same as the other comment:

not quite, if that's done and a new property is ever added to options and options does not contain useForceUpdate then that would break

e..g
```
useObserver(..., {
  someOtherOption: ...,
  // no useForceUpdate since it is optional
}

const forceUpdate = options.useForceUpdate() // useForceUpdate will be undefined 
```

src/observer.ts
xaviergonz: removed from observer

src/useObserver.ts
FredyC: That's why I was suggesting default in previous comment :)

src/useObserver.ts
FredyC: Correction...This way it either takes one specified by the user or imported one if undefined.
```
options: IUseObserverOptions = {
  useForceUpdate = useForceUpdate
}
```

xaviergonz: That doesn't compile :(

FredyC: Ah nevermind, just realized what you mean. Let's do this at least so there is always an object for options.

```
export function useObserver<T>(
    fn: () => T,
    baseComponentName: string = "observed",
    options: IUseObserverOptions = {}
): T {
    if (isUsingStaticRendering()) {
        return fn()
    }

    const wantedForceUpdateHook = options.useForceUpdate || useForceUpdate
```

xaviergonz: done, but used an EMPTY_OBJECT constant so it won't waste time creating an empty object when there are no options (which is the usual)

xaviergonz:
Thanks!

FredyC:
Released 1.2.0

PS: Do you have a repo for that project building on top of this? I might be curious to have a look eventually :)

xaviergonz:
https://github.com/xaviergonz/mobx-react-component

Right now there's only the "class" syntax (so you might not like it :) ), but the building block hooks are there already, so it might be a matter of ironing them out.

Either way let me know what you think!

xaviergonz:
@FredyC  now with hooks support

sample code sandbox: https://codesandbox.io/s/jlmx7jk779

FredyC:
@xaviergonz It's definitely an interesting approach. I am more vested into the #109 right now as it's closer to my heart.

Issue #97, opened 2019-03-23
[Discuss] useObserver, utility hooks


joenoon:
I'm doing some experimentation/refactoring in light of #38 and #94, and just in general catching up.  I have a few places where useComputed works quite well.  What I'm starting to think is the pattern of `useObserver` as the last return is actually the root of a few problems.

For one, it means you can never do an early return (since a hook is the last return), which is a huge annoyance to me.

Second, using the `useObserver` hook as the last return creates a split in the component of unobserved/observed behavior, which I think is just adding complexity.  

I've started going back to wrapping components with `observer` directly and liking how it feels so far.  Under the hood it's still using `useObserver`, but it's around the entire component.

Happy to be completely wrong on this or have my mind changed, but is it possible `useObserver` is really something that should be more of an internal implementation detail and not exposed to the user?

FredyC:
I've been struggling with this for some time as well. But I've also realized it's simply possible to do the following.

```js
const DataFetchingComponent = ({ obs }) => {
  return useObserver(() => {
    const { data, loading } = useQuery({ variables: { value: obs.input })
    return <UiWithDate data={data} />
  })
}
```

It might feel strange at first, but it doesn't break any rules of hooks in the end. The `useObserver` does not need to be the last, it's just it needs to wrap around anything you need to observe. It's fairly similar to using `observer` HOC, but this way your tree stays clean.

joenoon:
I was doing that in a few places as well for the same reasons but I stopped.  Are you sure it doesn't break the rules of hooks?

When I add `'react-hooks/rules-of-hooks': 'error'` to my eslint rules I get "React Hook "useObservable" cannot be called inside a callback. React Hooks must be called in a React function component or a custom React Hook function."

I turned the rule off for a while just to use the wrap-everything useObserver style, but now I'm trying to turn it back on.  If it's valid, maybe this was a eslint rule bug at some point and I just need to update my libs.  I will give that a shot.

FredyC:
Well, I think it's a false positive from that rule considering it's actually working in runtime :) Sadly, there is probably no clean way for it to recognize that case automatically. Even though it's a callback, it will be synchronously executed always. I suppose that rule would need to support a configuration for hook names that would be excluded from that rule.

joenoon:
So a little more digging and it seems:

using `observer` - the eslint hook rules silently don't check these... which makes it tricky because I didn't know until I intentionally broke some rules.

[...]

PR #109, opened 2019-03-28
macro: createUseObserverMacro

joenoon:
docs are currently in the source only. tests also show setup.
this probably requires an addition to the npm release
process to include the /macro/ build output directory.

Per #97


joenoon:
Some additional notes:

- I chose to do this without adding/changing anything about mobx-react-lite's main bundle.  I think there are some benefits to this approach, especially at this stage.

- the `createUseObserverMacro` is a dev/build time construct, so it doesn't make sense in the main bundle.  I'm not sure if rollup should/could be used anyway to create the macro bundle, but if so let me know.

- By exposing the macro "creator" and no implementation either another package or the application can pretty easily create their own x.macro.js file and implementation (which could be as simple as 10 lines or so).  This way people can add their own `useX` hooks for mobx, decide on how to enforce usage, etc.  I think this is good in light of #94.  For now, the tests show how to set up an implementation, and maybe I'll put together an npm package for my implementation - although as its still evolving, it might be too early.

FredyC:
From what I know, the usual approach about macros is to be able `import { createUseObserver } from 'mobx-react-lite/macro`. Rollup would help with bundling more files into a single one, but the way you have done should be fine as well.

I would probably drop the "Macro" suffix since it's already in the imported path and it's easy to rename import if someone wants to be more explicit.

---

I am not exactly sold by this approach yet. The outcome has an interesting appeal for sure. But it feels somewhat ceremonial - that need to have is spread over 3 files (in case of TS). Chances are that most people would not want to go through that boilerplate. Not even mentioning if you need to do it several times. What if you have a somewhat modular application and each part includes its own Context-based observable store you want to include in that macro?

joenoon:
> But it feels somewhat ceremonial - that need to have is spread over 3 files (in case of TS). Chances are that most people would not want to go through that boilerplate. Not even mentioning if you need to do it several times. What if you have a somewhat modular application and each part includes its own Context-based observable store you want to include in that macro?

Perhaps, but in light of #94 and still being early in the process of figuring out the best way forward, providing the tools for people to do this fairly easily in a very flexible way seems appropriate.  I'll take the 3 files over the boilerplate/ceremony without it across 50+ application files any day! :)

I'm not sure of the need to do it several times.  One time should be able provide any custom hook functions regardless of context in the normal way hooks work.  But maybe there is a case I'm not seeing?

FredyC:
> I'm not sure of the need to do it several times. One time should be able provide any custom hook functions regardless of context in the normal way hooks work. But maybe there is a case I'm not seeing?

As I said, a modular architecture might want to have a separate context for each module and consolidating them in a single file for the macro could be an obstacle.

Either way, I need to try this out to get some feeling for it.

joenoon:
@FredyC A couple very different thoughts are coming to mind:

Originally, it was the usage of useObservable/useComputed _outside_ of a useObserver that I was most concerned with preventing.  We could easily add a "default" macro similar to the one in the tests that provides this, but that sort of goes against #94 and creates a weird feeling when the developer would like their own `useReaction` or whatever to have a similar feel.

We discussed the idea of passing something, like an isObserved, Symbol, or the provided hooks themselves as a sort of "guard" for user-created hooks that need useObservable, useComputed, etc.

Then I realized we actually don't need a guard since those hooks could simply call `useObserver()` just like a component would.  This would work but is probably unnecessary.

BUT THEN I thought, wait, maybe we don't need to do a lot of this.  The macro already does the ugly part for us (IMO much cleaner than Observer, observer, or useObserver(fn)).

If we JUST exposed the macro that wraps the function for you like useObserver works today, people can import useObservable like they do today, other hooks they've created and so on, and use it exactly like they do today.  And if those utility hooks go away at some point they can import them from their own code.

I think we only need one rule of thumb for this to work: if you expect your component to be observed or use hooks that expect it, you have to call `useObserver()` in the _Component_ (not your own custom hooks).

This seems like the exact same rule with traditional mobx-react's observer, just with a different style.

So, how does this sound? :
- Expose a real implementation of the macro so the user doesn't have to set anything up.
- `useObserver()` will return void
- Users import their useObservable, useComputed, etc. as normal
- We simply provide an example showing proper component structure (we can make the example better):

```
import { useObserver } from "mobx-react-lite/macro"
import { useObservable, useComputed } from "mobx-react-lite"

export const MyComponent = props => {
    useObserver()
    const foo = useObservable({ foo: 1 })
    const bar = useComputed(() => foo.foo + 1)
    return "Hello, world!"
}
```

I didn't quite follow your module architecture (is that separate apps, or just different areas of the same app?), but I'm curious if this would go toward solving that or if you have any ideas on this direction?

joenoon:
Pushed this simpler (and the more I think about it, better) version!  With a few bugfixes as well: one AST issue and previously was ignoring the args that can be passed to `useObserver` after the fn.

FredyC:
I kinda do like that as an effortless way, no ceremony required. But it does not really solve the essential problem that you have to keep remembering to `useObserver` there otherwise things won't work.

It is easier for linter, it's less verbose and it's slightly more apparent than regular `observer` HOC, but that's about it. It might feel a bit strange considering it's a hook that does not accept or return any values. One might think it does some _magical_ side effect. Perhaps it should be called something like `applyObserver` instead to denote that difference.

There are also some edge cases to solve. What if you have multiple `useObserver` calls there? Does the order matter? I would at least add invariant that you have to have `useObserver` on the first line of component code (if that's detectable). Without a clear way specification where to use that, people would have it scattered all over different places.

---

Considering all that, I am definitely more inclined to a first version even over the ceremony. Sadly, I am a bit swamped with other tasks right now, so I cannot provide any promises when I get to try this out.

Don't mind the "modular architecture", in most cases it's just a wet dream to have things in the app cleanly modularized, but it's purely scientific.

joenoon:
@FredyC I would like to get this in unless you see some required changes needed.  I summed up a good bit of the reasoning behind this approach here: https://github.com/mobxjs/mobx-react-lite/issues/94#issuecomment-479115437 .  As for `applyObserver` instead of a `use...` name, I'm not sure if I feel strongly one way or another.  Because whatever we call it actually executes a hook (the real `useObserver`), my first reaction is it should have a `use...` name.   But I guess it would probably "work" either way.  If you would prefer this as a separate library that seems workable as well, although for adoption sake and getting to "one clear way", I think it would be much better in this library.

FredyC:
> Because whatever we call it actually executes a hook

That's true. It's only a matter of semantics. Point being that it's feels weird to have a hook that seemingly does nothing. Someone who does not know the "magic" behind would be scratching ahead how does that work. With `applyObserver` it stands out like something else.

Honestly, I am not too convinced about the approach as it adds slight weirdness into the code. It feels like macro abuse. It doesn't directly transform some code on input but wraps the code of the function body. That's fairly unusual for a macro in my opinion.

> If you would prefer this as a separate library that seems workable as well, although for adoption sake and getting to "one clear way", I think it would be much better in this library.

I am conflicted on this. It's an alternative approach that has nothing to do with MobX itself. But it's not that unusual to have macro exported from the package like this. At least it would deserve some better documentation and reasoning behind it. From the small example in README it's clear why would anyone use that. You could explain there what you said in https://github.com/mobxjs/mobx-react-lite/issues/94#issuecomment-479115437

Otherwise in a matter of code review, I don't have any objections.

joenoon:
@FredyC Great, I've added more to the README from that comment and rebased.  I agree there is plenty of opportunity for discussion around where we want to go from here so I've moved the macro docs outside of the API docs to give it some separation.  (side note: vscode reformatted the table of contents on Save.  I think its the expectation given the included vscode settings, but I guess it could be something coming from my own settings.  Let me know, easy to fix if its a problem)

FredyC:
Ok, I suppose docs are ok like that. Some response to my `applyObserver` concerns?

joenoon:
Yes, `applyObserver` vs `useObserver`:
- Naming: for me it feels like a hook, actually rewrites to a hook, so should be named like a hook
- Seemingly does nothing/magic:  I guess I could see some of that perspective, but I see the macro version as the equal to `observer` everyone is used to already, which has many of those parallels.  It's also not much different from extracting something into a custom `useEffect` hook that you call without arguments or expecting a return.

So I see this pattern lining up well with the progression of mobx into the hooks world as well as following hooks style in general.  It's kind of like starting from the end and working backwards: What do we wish we could write in an ideal world?  Then given the limitations, can we transpile our way there so we actually can write it that way today?  There is plenty of this from decorators, generators, js in general, etc to point to.

FredyC:
> * but I see the macro version as the equal to `observer`

Sorry, that's hardly relevant if we are deciding how to name the macro.

Among the mentioned issues, I see the problem of adoption. Consider there are some code bases that already have `useObserver` all over the place. If they want to switch to macro, they have to analyze the import in that file and rewrite it. It's super confusing to see the same name which behaves differently based on a slight difference in import.

Also, auto-import in VSCode will be offering two `useObserver`. If you make a mistake, import runtime one and try to use it as a macro, you will only find out about the problem in runtime and that's really bad.

In summary, I don't think I want to merge it named like that. So if you insist, I guess it will have to live in a separate package.

joenoon:
I'm going to close this in preparation for a separate library.  I think the name confusion, two auto-import suggestions as you mentioned, etc. would have been a problem.  Plus just the extra confusion of a 4th way of doing things.

FredyC:
Interesting, so you think it would be a problem and yet you are going to have `useObserver` in the same lib? It will suffer exactly the same problems :)

joenoon:
Well not exactly.  I'm removing quite a bit of stuff and keeping the lib strictly focused on hooks, having only one way to make a component observed (`useObserver`), a modified useObservable I've been using, and the current useComputed.  The runtime hook will be called `useObserverInternal` and not have typedefs, so will not show up.  In the end we all want to get to the same place, but adding yet another way of doing things to a library might not be the best way to get there.  If it doesn't work out I'll be back, of course! 👍

PR #120, opened 2019-04-03
Schedule uncommitted reaction cleanup

RoystonS:
This PR builds on PR #119 and introduces some fixes for #53:

- a unit test for checking that reactions are properly disposed for components that were double-rendered before being committed (e.g. by ConcurrentMode or StrictMode).
- a fix which schedules the cleanup of such reactions.

Note that there's one aspect I'm unsure of, both of how to test or fix in a nice way, so I'm pushing this up for feedback.  cc: @mweststrate, @FredyC

The happy day case, which I've tested for, and coded for:

1. render phase 1:
    1. Component instance A gets rendered, creating reaction R1, which we track
    1. we trigger a timer to check for leaked reactions
    1. rendering is aborted and restarts, or StrictMode triggers a second render in order to root out trouble
    1. Component instance A gets rendered, creating reaction R2, which we track
1. commit phase 1:
    1. Component instance A commits, and we remove R2 from the 'leaked reactions' list
1. render phase 2:
    1. Component instance B gets rendered, creating reaction R3, which we track
    1. second render triggered
    1. Component instance B gets rendered, creating reaction R4, which we track
1. commit phase 2:
    1. Component instance B commits, and we remove R4 from the 'leaked reactions' list
1. cleanup timer ticks:
    1. R1 and R3 are disposed. All is well. Everybody is happy.

But there's an edge case I'm concerned about (which is why I don't think that this PR is sufficient). What if it's been a while since the timer was kicked off, but _the latest render phase hasn't had a corresponding commit phase yet_?

1. render phase 1:
    1. Component instance A gets rendered, creating reaction R1, which we track
    1. we trigger a timer to check for leaked reactions
    1. rendering is aborted and restarts, or StrictMode triggers a second render in order to root out trouble
    1. Component instance A gets rendered, creating reaction R2, which we track
1. commit phase 1:
    1. Component instance A commits, and we remove R2 from the 'leaked reactions' list
1. render phase 2:
    1. Component instance B gets rendered, creating reaction R3, which we track
    1. second render triggered
    1. Component instance B gets rendered, creating reaction R4, which we track
1. cleanup timer ticks (between render and commit - this _can_ happen, right?) :
    1. R1, R3 and R4 are disposed
1. commit phase 2:
    1. Component instance A commits, and the world ends because R4 has previously been disposed by the cleanup schedule

I _think_, to do this properly, we need to record that a reaction is safe to clean up only once a commit phase has happened?

### DIFF
--- a/src/useObserver.ts
+++ b/src/useObserver.ts
@@ -1,8 +1,8 @@
 import { Reaction } from "mobx"
-import { useDebugValue, useRef } from "react"
+import { useDebugValue, useEffect, useRef } from "react"
 import { printDebugValue } from "./printDebugValue"
 import { isUsingStaticRendering } from "./staticRendering"
-import { useForceUpdate, useUnmount } from "./utils"
+import { useForceUpdate } from "./utils"
 
 export type ForceUpdateHook = () => () => void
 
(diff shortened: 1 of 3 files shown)


RoystonS:
(Coverage drop is due to removal of last use of `onUnmount`, as mentioned in #119. Can remove if desired.)

FredyC:
Wow, this is going to take some time to review for sure :) Also I am slightly confused. Should we merge #119 first? Why two separate PRs?

RoystonS:
Yes, I think #119 could be merged separately, and there may well be value in doing that because, whilst it doesn't solve the leak, it does quieten down StrictMode by avoiding the unnecessary state mutations, making it more useable.

I think the reaction cleanup will take a little longer to work through, and that doesn't need to hold up #119.

FredyC:
Can you please resolve conflicts to see the relevant changes only?

RoystonS:
I've replaced this with PR #121  as I had some issues with my repo and then couldn't push to PR #120.

Issue #129, opened 2019-04-15
Using state variable inside reaction always return undefined


MatthieuCoelho:
I'm creating variable in state : 
const [answerId, setAnswerId] = useState<boolean | undefined>();

Then I set it :
setAnswerId(false)

That I use inside a reaction
```
useDisposable(() => reaction(
        () => /*Random observable*/,
        (myObservable) => {
           console.log(answerId) //Always show undefined
        }
   }));
```

Shouldn't it works ?

mweststrate:
No, useState doesn't create observables, so reactions won't react to them.
Instead, use `useEffect` instead if you want to react to useState created
things, or create some local observables, like: `const [store] =
useState(() => observable({ answerId: undefined }))`. See also #94

MatthieuCoelho:
I'm agree with that, the trigger of the effect is a real observable for outside ...

Shouldn't the state be accessible inside the reaction ?

mweststrate:
It is, but due to how javascript closures work, you will only see the first
value, which is probably what you are interested in, but technically correct

Op di 16 apr. 2019 01:12 schreef MatthieuC <notifications@github.com>:

MatthieuCoelho:
@mweststrate thanks, so the only way to "fix" it is to use a local observable otherwise I won't me able to get the current state value.

mweststrate:
There are several ways to fix it, such as using useEffect, instead of
useDisposable (which is going to be removed soon). However, we are still in
the progress of standardizing the _idiomatic_ ways to achieve this. See
also #97 and the open PR

Issue #106, opened 2019-03-26
useComputed outside useObserver is uncached


xaviergonz:
since useComputed returns a plain value, this means that if it is used outside a useObserver block it will never cache, and therefore just be a more expensive "always calculate this"

e.g.
```
const sum = useComputed(() => 1+2); // this will return the plain value 3, and since get() is called outside observation it is not cache
useObserver(() => {
  // if we use sum here it will just use the plain value 3, which cannot be tracked
})
```

xaviergonz:
One way to fix it I guess would be to return a
```
{
  get current() { return computed.get() } // readonly
}
```

and then use it like
```
const sum = useComputed(() => 1+2)
useObserver(() => {
  // if we use sum.current here it will actually call get, which can be tracked
})
```

FredyC:
I wouldn't bother really since we are considering removal of it ... #94

FredyC:
Closing since we have indeed deprecated `useComputed` in favor of `useLocalStore`. Released in 1.3.0

Issue #22, opened 2018-11-18
Idea: Add useAction hook?


FredyC:
Looking at the example in README, I have a feeling that this is going to be rather a repeated pattern.

```ts
    const todos = useObservable(new Map<string, boolean>())
    const addTodo = React.useCallback(() => {
        todos.set(todoRef.current.value, false)
        todoRef.current.value = ""
    }, [])
    const toggleTodo = React.useCallback((todo: string) => {
        todos.set(todo, !todos.get(todo))
    }, [])
```

So I am thinking something along the lines of another tiny wrapper around the `useCallback` with empty inputs array by default. Possibly it can use `runInAction` internally. Thoughts?

@xaviergonz @mweststrate

xaviergonz:
```
const actions = useActions({
  addTodo() {
    ...
  },
  toggleTodo(todo: string) {
    ...
  }
})
...
actions.addTodo();
```
With each action in the object wrapped in action with the name set to the object key?

xaviergonz:
What would be the case where a non empty array of inputs would be useful?

xaviergonz:
Then again, this is kind of supported by observable decorators, so this could work as well?

```js
const person = useObservable({
    name: "John",
    age: 42,
    showAge: false,

    get labelText() {
        return this.showAge ? `${this.name} (age: ${this.age})` : this.name;
    },

    // action:
    setAge(age) {
        this.age = age;
    }
}, {
    setAge: action
    // the other properties will default to observables  / computed
});
```

FredyC:
Ah, I thought it makes actions automatically from any functions in that object. In that case, there is a mistake in README as well :)

It feels somewhat nicer to have a separate hook for defining actions. This syntax is somewhat confusing if you need to list all actions in a second argument.

Also, can you actually use the Map inside the observable object? Haven't tried that before.

xaviergonz:
Sure, you can even do this I guess

```js
class Todo {
  ...
  @action toggle() {...}
}
class TodoStore {
  todos = new Map<string, Todo>();
  @action addTodo(): Todo { ...}
}
...
const store = useObservable(new TodoStore); 
store.addTodo()
store.todos[id].toggle();
etc
```

FredyC:
Meh. Classes 🙄 I like [MST](https://github.com/mobxjs/mobx-state-tree) more, but that's not the point. We are not talking about observable outside of the component, but inside to replace a more verbose `useReducer` (I really dislike those strings for action types). 

[...]

PR #130, opened 2019-04-15
Implement new hooks proposal, as discussed in #94

mweststrate:
* [x] make sure `useLocalStore` automatically converts functions to actions and binds them (and update readme example)?
* [x] make sure `useAsObservableSource` only accepts plain objects (not arrays)
* [x] tests
* [x] implementation
* [x] investigate if prop changes introduce double renders 
* [x] forbid (or support) changes in the observable property set
* [x] update TOC
* [ ] ~remove old docs~
* [ ] ~remove old hooks~
* [ ] ~fix double rendering~

Update issues #129, #106, #97, #94, #38, #22 accordingly after release


README.md
```diff
@@ -142,6 +142,135 @@ const Person = memo(props => {
...
+    ))
+}
+```
+
+In the above example, any change to `multiplier` prop will show up in the `observableProps` observable object, and be picked up by the `store`.
+
+Reading properties for the returned observable object, should always be done _inside_ the store initializer (respectively side-effect, see below), to avoid values being trapped in the closure.
+In other words: _the return value of `useAsObservableSource` should never be deconstructed! So, don't write: `const {multiplier} = useAsObservableSource({ multiplier })`!_
+
+Note that the value passed to `useAsObservableSource` should always be an object, and is made only shallowly observable.
```
FredyC: This is fairly important to be just a note imo. Honestly, I am unsure myself what would happen with array value on that passed object. Is it going to be observable on itself? I am a bit 🤧 so I not thinking clearly right now...

README.md
```diff
@@ -142,6 +142,135 @@ const Person = memo(props => {
...
+
+The object returned by `useAsObservableSource`, although observable, should be considered read-only for all practical purposes.
```
FredyC: Can't we use the functionality of MST to actually prevent changing values directly? Not sure if it's part of MobX or not.

README.md
```diff
@@ -142,6 +142,135 @@ const Person = memo(props => {
...
+    return useObserver(() => (
+        <div>
+            Count: {store.count}
+            <button onClick={store.inc}>Increment</button>
+        </div>
+    ))
+}
+```
+
+It is important to realize that the store is created only once! It is not possible to specify dependencies to force re-creation, _nor should you directly be referring to props for the initializer function_, as changes in those won't propagate.
```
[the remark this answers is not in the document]
mweststrate: As you said before, throwing away a store is quite impactful, especially as you might not want to reset it entirely. I think the current solution with `asObservableSource` is better.

README.md
```diff
@@ -142,6 +142,135 @@ const Person = memo(props => {
...
+            Multiplied count: {store.multiplied}
+            <button onClick={store.inc}>Increment</button>
+        </div>
+    ))
+}
+```
+
+In the above example, any change to `multiplier` prop will show up in the `observableProps` observable object, and be picked up by the `store`.
+
+Reading properties for the returned observable object, should always be done _inside_ the store initializer (respectively side-effect, see below), to avoid values being trapped in the closure.
```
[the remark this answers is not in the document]
mweststrate: Maybe we can just drop it, I think the docs are reiterating the same thing a few times now

README.md
mweststrate: I think it can be a note, because in all other case we will be throwing an exception immediately anyway :P

README.md
mweststrate: I was thinking about maybe return an object with just getters, that wraps the real one. But might be over-engineering at this point.

src/useAsObservableSource.ts
```diff
@@ -0,0 +1,17 @@
...
+import { useState } from "react"
+
+import { isPlainObject } from "./utils"
+
+export function useAsObservableSource<T>(current: T): T {
+    if (!isPlainObject(current)) {
+        throw new Error("useAsObservableSource expects an object as first argument")
+    }
+
+    const res = useState(() => observable(current, {}, { deep: false }))[0]
```
FredyC: Why not like that? I mean that `[0]` is kinda hidden there and I was confused at first. What's the benefit?
```suggestion
    const [res] = useState(() => observable(current, {}, { deep: false }))
```

src/useAsObservableSource.ts
```diff
@@ -0,0 +1,17 @@
...
+    if (Object.keys(res).length !== Object.keys(current).length) {
```
FredyC: We should wrap this into a NODE_ENV check just to make it slightly faster in prod.

README.md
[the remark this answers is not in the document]
FredyC: Yea, but I was thinking something along the lines the component is rendering for some ID and then you receive a different ID. The reset of the state might be desired in that case. But we can surely add that later if it proves to have some actual use case.

README.md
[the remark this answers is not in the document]
mweststrate: updated

src/useAsObservableSource.ts
mweststrate: Done

src/useAsObservableSource.ts
mweststrate: Done

README.md
```diff
@@ -18,10 +18,9 @@ Project is written in TypeScript and provides type safety out of the box. No Flo
     -   [`<Observer/>`](#observer)
     -   [`observer<P>(baseComponent: FunctionComponent<P>, options?: IObserverOptions): FunctionComponent<P>`](#observerpbasecomponent-functioncomponentp-options-iobserveroptions-functioncomponentp)
     -   [`useObserver<T>(fn: () => T, baseComponentName = "observed", options?: IUseObserverOptions): T`](#useobservertfn---t-basecomponentname--%22observed%22-options-iuseobserveroptions-t)
-    -   [`useObservable<T>(initialValue: T): T`](#useobservabletinitialvalue-t-t)
```
FredyC: I am thinking we shouldn't be probably removing those hooks in this PR. They should become deprecated and released in 1.X with those new hooks. Then the 2.X can come out with those removed. 

Personally, I don't want to be refactoring whole apps right away :)

mweststrate: Ok, reverted, added warning instead

src/useComputed.ts
```diff
@@ -1,7 +1,14 @@
 import * as mobx from "mobx"
 import { useMemo } from "react"
 
+let warned = false
+
 export function useComputed<T>(func: () => T, inputs: ReadonlyArray<any> = []): T {
+    if (process.env.NODE_ENV !== "production" && !warned) {
+        warned = true
+        // tslint:disable-next-line: no-console
+        console.warn("[mobx-react-lite] useComputed has been deprecated")
```
FredyC:
```suggestion
        console.warn("[mobx-react-lite] useComputed has been deprecated. Use useLocalStore instead.")
```

src/useDisposable.ts
```diff
@@ -20,6 +22,11 @@ export function useDisposable<D extends TDisposable>(
     disposerGenerator: () => D,
     inputs: ReadonlyArray<any> = []
 ): D {
+    if (process.env.NODE_ENV !== "production" && !warned) {
+        warned = true
+        // tslint:disable-next-line: no-console
+        console.warn("[mobx-react-lite] useDisposable has been deprecated. Use useEffect instead.")
```
FredyC:
```suggestion
        console.warn("[mobx-react-lite] useDisposable has been deprecated. Use React.useEffect instead.")
```

src/useObservable.ts
```diff
@@ -3,7 +3,17 @@ import { useRef } from "react"
...
 type SupportedValues = object | Map<any, any> | any[]
 
+let warned = false
+
 export function useObservable<T extends SupportedValues>(initialValue: T): T {
+    if (process.env.NODE_ENV !== "production" && !warned) {
+        warned = true
+        // tslint:disable-next-line: no-console
+        console.warn(
+            "[mobx-react-lite] useObservable has been deprecated. Use useLocalStore instead"
```
FredyC:
```suggestion
            "[mobx-react-lite] useObservable has been deprecated. Use useLocalStore instead."
```

src/useComputed.ts
FredyC: So you don't agree with this suggestion? I mean there should be some recommendation when deprecating something.

FredyC:
Just want to make sure. Are you aware of linter failing? I would fix it, but I don't want to cause unnecessary merge conflicts and such. Hard to say if tests are actually passing :)

I guess double rendering cannot be really fixed after all?

mweststrate:
ah, I fixed those, but forgot to push indeed :)

mweststrate:
Fixed

FredyC:
LGTM.

Would you mind adding simple tests to cover those error scenarios? It's an unnecessary drop in coverage imo.

https://coveralls.io/builds/22890236/source?filename=src%2FuseAsObservableSource.ts#L8

I am not entirely sure why there is a drop in `printDebugValue` as no related code should be changed about that, but I suppose it will be resolved in #121 anyway, so I wouldn't bother much.

README.md
```diff
@@ -142,12 +145,84 @@ const Person = memo(props => {
...
+```
+
+In the above example, any change to `multiplier` prop will show up in the `observableProps` observable object, and be picked up by the `store`.
+
+Warning: _the return value of `useAsObservableSource` should never be deconstructed! So, don't write: `const {multiplier} = useAsObservableSource({ multiplier })`!_
+
+The value passed to `useAsObservableSource` should always be an object, and is made only shallowly observable.
+
+The object returned by `useAsObservableSource`, although observable, should be considered read-only for all practical purposes.
+Use `useStore` to create local, observable, mutable, state.
```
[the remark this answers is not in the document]
FredyC: Can you please use the suggestion feature next time? It allows to commit it right away.
```suggestion
Use `useLocalStore` to create local, observable, mutable, state.
```
image

README.md
```diff
@@ -142,12 +145,84 @@ const Person = memo(props => {
...
+
 # Notice of deprecation
 
 We are considering deprecation and removal of following utilities from the package. Come join the discussion: https://github.com/mobxjs/mobx-react-lite/issues/94
```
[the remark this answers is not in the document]
FredyC: Yea, that should be removed, good catch.

README.md
```diff
@@ -266,6 +345,64 @@ const Name = observer((props: { firstName: string; lastName: string }) => {
...
+
+    return /* etc */
+}
+```
+
+Note that the disposer function of `autorun` should be returned to `useEffect` so that the effect is cleaned up properly by React.
+
+Secondly, when using MobX based side effects, you typically don't want to re-create them after each rendering, so make sure to pass in an empty array `[]` as deps to `useEffect`.
+
+This will yield the same limitation as when using `useStore`: changes to props used by the side-effect won't be picked up automatically, so don't refer to them directly. Instead, leverage `useAsObservableSource` again:
```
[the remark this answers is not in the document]
FredyC:
```suggestion
This will yield the same limitation as when using `useLocalStore`: changes to props used by the side-effect won't be picked up automatically, so don't refer to them directly. Instead, leverage `useAsObservableSource` again:
```

README.md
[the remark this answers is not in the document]
rvion: of course ! thanks for the tip !

src/useLocalStore.ts
```diff
@@ -0,0 +1,18 @@
...
+import { isPlainObject } from "./utils"
+
+export function useLocalStore<T>(initializer: () => T): T {
+    return useState(() => {
+        const store: any = observable(initializer())
+        if (isPlainObject(store)) {
+            Object.keys(store).forEach(key => {
+                const value = store[key]
+                if (typeof value === "function") {
+                    store[key] = action(value.bind(store))
```
mweststrate: Probably we should only bind, but not `action`, as it will kill derivations that use local functions to compute things

FredyC:
Oh. I've messed up slightly as I forgot this is still going to V1. Feel free to force push next time.

src/useLocalStore.ts
FredyC: I am not entirely sure I understand the problem here, but hey I never properly understood the exact action mechanics. Let's merge it like this and if it turns out to be a problem, we can remove it.

Issue #138, opened 2019-04-29
Going from mobx-react -> mobx-react-lite


johnnyreilly:
Heya!

A while ago I took mobx-react-lite for a spin.  The problem I bumped was how I should migrate the following pattern to mobx-react-lite:

```
interface IInjectedProps {
    login: () => void;
    isAuthenticated: boolean;
}

type IProps = RouteComponentProps & Partial<IInjectedProps>;

export const HomePageRaw: React.FC<IProps> = props => {
        const { login, isAuthenticated  } = props;

        return (
            <div>html be here</div>
        );
    }

export const HomePage = inject(
    (dependencies: Dependencies): IInjectedProps => ({
        login: dependencies.authStore.login,
        isAuthenticated: dependencies.authStore.isAuthenticated
    })
)(observer(HomePageRaw));
```

I tried about 5 or so permutations based on the docs but things generally either didn't work or didn't work as expected (perpetual re-renders etc)

Given that this is a fairly common pattern, I wonder if it would be possible to request that an example be added to the docs illustrating how you would go from this to a mobx-react-lite friendly alternative.

From what I understand `inject` is dead in a ditch, instead reach for `useContext` - that's cool.  But how should the rest of the component be structured?  I'd be happy to send a docs PR if I understood how things could / should be structured. 

If someone could advise I'd be super grateful.

Thanks for all the work you're doing BTW.  It's 💯 😄

FredyC:
Um, can you perhaps show at least one pattern you have tried and failed with it? I cannot imagine where what's the catch really :)

joenoon:
FWIW, this example from the tests (with a couple changes to make sure eslint is happy) probably covers 98+% of the use cases.  I've yet to switch over to the new useLocalStore/useAsObservableSource hotness, but trust the tests.  Essentially, every component with observable data should look something like this:

```tsx
import { observer, useAsObservableSource, useLocalStore } from "mobx-react-lite"

interface CounterProps {
  multiplier: number;
}

export const Counter = observer(function Counter(props: CounterProps) {
    const observableProps = useAsObservableSource(props)
    const store = useLocalStore(() => ({
        count: 10,
        get multiplied() {
            return observableProps.multiplier * this.count
        },
        inc() {
            this.count += 1
        }
    }))

    return (
        <div>
            Multiplied count: <span>{store.multiplied}</span>
            <button id="inc" onClick={store.inc}>
                Increment
            </button>
        </div>
    )
})
```

[...]

Issue #136, opened 2019-04-29
Deprecated useObservable is used in the first example in README.md


shamrin:
It's confusing that useObservable is in the very first example in README.md. And then it's mentioned that useObservable is deprecated. Could you please fix the examples so they don't use deprecated API?

FredyC:
I would accept PR :) I am not available till the end of the week.

shamrin:
Thank you @FredyC! Unfortunately, I'm very new to mobx-react-lite. What do I use instead of `useObservable`?

FredyC:
If you like, please check out the discussion leading to a deprecating: #94

mweststrate:
See #140

Ciantic:
I'm here to say that I'm trying to figure out how to replace my useObservable calls. But the documentation is truly confusing. I think merging that change of docs would be most helpful for everybody.

ricardobeat:
Concur with @Ciantic above. I'm looking for a new solution for managing state in a fully hooks-based project; after reading through #94, #130, #140, other issues, mobxjs/mobx-react#640 I still have little clue on the current state of things.

- All four initial examples in the README contain `useObservable()`
- The last item in the README shows how to use `React.Context` but doesn't tell us what `createStore()` is
- In mobxjs/mobx-react#640 you'll find this comment:

> V6 does not really bring any new essential features on the table. It's pretty much a starting point where class and functional (with hooks) components are supported in a single package. It uses mobx-react-lite underneath

but no mention of hooks at all in the main package docs (it points back here instead)
- The [last example in this post](https://dev.to/ryands17/using-mobx-with-react-hooks-52h5) is a bit more clear but uses a mix of `mobx` and `mobx-react-lite` which I'm now unsure is the right approach

Could somebody who's been following development explain, or maybe point to an updated canonical example? Should we just hold off on MobX + hooks for the meantime?

FredyC:
I understand it's not ideal, but we are simply lacking time to finish up #140. Just read through that for now, please. 

https://github.com/mobxjs/mobx-react-lite/blob/readme-improvements/README.md

> All four initial examples in the README contain useObservable()

The `useObservable` is clearly marked as deprecated there.

> The last item in the README shows how to use React.Context but doesn't tell us what createStore() is

That's out of scope and up to you how you construct your store, eg. class+decorators or mobx-state-tree

[...]

PR #140, opened 2019-04-30
some readme improvements

mweststrate:
Addresses #138, #136

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,10 @@ Class based components **are not supported** except using `<Observer>` directly 
 
 Project is written in TypeScript and provides type safety out of the box. No Flow Type support is planned at this moment, but feel free to contribute.
 
+-   [Overview](#overview)
+    -   [Adding reactivity to React components](#adding-reactivity-to-react-components)
+    -   [Introducing local observable state](#introducing-local-observable-state)
+    -   [More extensive example](#more-extensive-example)
 -   [API documentation](#api-documentation)
     -   [`<Observer/>`](#observer)
     -   [`observer<P>(baseComponent: FunctionComponent<P>, options?: IObserverOptions): FunctionComponent<P>`](#observerpbasecomponent-functioncomponentp-options-iobserveroptions-functioncomponentp)
@@ -28,6 +32,121 @@ Project is written in TypeScript and provides type safety out of the box. No Flo
 -   [Server Side Rendering with `useStaticRendering`](#server-side-rendering-with-usestaticrendering)
 -   [Why no Provider/inject?](#why-no-providerinject)
 
+## Overview
+
+This library has two goals:
+
+1. Make it possible to introduce reactivity to React function components
+2. Make it possible to have local observable state in React function components
+
+This library does _not_ support class based components. If you need class based component support, use [mobx-react](https://github.com/mobxjs/mobx-react), which wraps this library.
+
+### Adding reactivity to React components
+
+There are three ways to add reactivity to components: `observer(fn)`, `<Observer>{renderFn}</Observer>` and `useObserver(renderFn)`.
+In most cases, there difference doesn't matter that much, but here is an overview of the three different approaches:
+
+```javascript
+import { observable } from "mobx"
+import { Observer, useObserver, observer }  from "mobx-react-lite"
+
+const person = observable({
+    name: "John"
+})
+
+const P1 = observer(function P1({ person }) {
+    return <h1>{person.name}</h1>
+})
+
+const P2 = ({ person }) => (
+    <Observer>{() => <h1>{person.name}</h1>}</Observer>
+)
+
+const P3 = ({ person }) => {
+    return useObserver(() => (
+        <h1>{person.name}</h1>
+    ))
+})
+
+ReactDOM.render(<div>
+    <P1 person={person} />
+    <P2 person={person} />
+    <P3 person={person} />
+</div>)
+
+setTimeout(() => {
+    person.name = "Jane"
+}, 1000)
+```
+
+When using `observer`, the component `props` will be made automatically observable (so that they can be used in reactions and computed values), and `React.memo` will be applied automatically to the comment.
+
+`useObserver` keeps the component tree flat as it doesn't introduce a higher order component, but to make props observable, the [`useAsObservableSource`](#useasobservablesourcetstate-t-t) hook should be used.
+
+`<Observer>` is slightly better optimized compared to `useObserver`, but does introduce an additional component in the component tree.
+
+### Introducing local observable state
+
+Local observable state can be introduced by using the [`useLocalStore`](#uselocalstoretinitializer---t-t) hook, that runs once to create an observable store. A quick example would be:
+
+```javascript
+import { useLocalStore, useObserver } from "mobx-react-lite"
+
+const P3 = () => {
+    const todo = useLocalStore(() => ({
+        title: "Test",
+        done: true,
+        toggle() {
+            this.done = !this.done
+        }
+    }))
+
+    return useObserver(() => (
+        <h1 onClick={todo.toggle}>
+            {todo.title} {todo.done ? "[DONE]" : "[TODO]"}
+        </h1>
+    ))
+})
+```
+
+When using `useLocalStore`, all properties of the returned object will be made observable automatically, getters will be turned into computed properties, and methods will be bound to the store and apply mobx transactions automatically.
+
+_Note: using `useLocalStore` is mostly beneficial for really complex local state, or to obtain more uniform code base. Note that using a local store might conflict with future React features like concurrent rendering._
+
+### More extensive example
+
+The following example combines all concepts mentioned so far: `useLocalStore` to create a local store, and `useAsObservableProps` to make the props observable, so that it can be uses savely in `store.multiplied`:
+
+```typescript
+import { observer, useAsObservableSource, useLocalStore } from "mobx-react-lite"
+
+interface CounterProps {
+    multiplier: number
+}
+
+export const Counter = observer(function Counter(props: CounterProps) {
+    const observableProps = useAsObservableSource(props)
+    const store = useLocalStore(() => ({
+        count: 10,
+        get multiplied() {
+            return observableProps.multiplier * this.count
+        },
+        inc() {
+            this.count += 1
+        }
+    }))
+
+    return (
+        <div>
+            Multiplied count: <span>{store.multiplied}</span>
+            <button id="inc" onClick={store.inc}>
+                Increment
+            </button>
+        </div>
+    )
+})
+```
+
 ## API documentation
 
 ### `<Observer/>`
@@ -38,10 +157,10 @@ The rendering in the function will be tracked and automatically re-rendered when
 This can come in handy when needing to pass render function to external components (for example the React Native listview), or if you want to observe only relevant parts of the output for a performance reasons.
 
 ```jsx
-import { Observer, useObservable } from "mobx-react-lite"
+import { Observer, useLocalStore } from "mobx-react-lite"
 
 function ObservePerson(props) {
-    const person = useObservable({ name: "John" })
+    const person = useLocalStore(() => ({ name: "John" }))
     return (
         <div>
             {person.name}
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,6 +32,121 @@ Project is written in TypeScript and provides type safety out of the box. No Flo
...
+This library does _not_ support class based components. If you need class based component support, use [mobx-react](https://github.com/mobxjs/mobx-react), which wraps this library.
+
+### Adding reactivity to React components
+
+There are three ways to add reactivity to components: `observer(fn)`, `<Observer>{renderFn}</Observer>` and `useObserver(renderFn)`.
+In most cases, there difference doesn't matter that much, but here is an overview of the three different approaches:
+
+```javascript
+import { observable } from "mobx"
+import { Observer, useObserver, observer }  from "mobx-react-lite"
```
FredyC:
```suggestion
import { Observer, useObserver, observer }  from "mobx-react-lite"
import ReactDOM from "react-dom"
```

README.md
```diff
@@ -28,6 +32,121 @@ Project is written in TypeScript and provides type safety out of the box. No Flo
...
+    <P1 person={person} />
+    <P2 person={person} />
+    <P3 person={person} />
+</div>)
+
+setTimeout(() => {
+    person.name = "Jane"
+}, 1000)
+```
+
```
FredyC:
```suggestion
[![Edit MobX in React - observer pattern](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/4qrv105l47?fontsize=14)
```

README.md
```diff
@@ -28,6 +32,121 @@ Project is written in TypeScript and provides type safety out of the box. No Flo
...
+When using `observer`, the component `props` will be made automatically observable (so that they can be used in reactions and computed values), and `React.memo` will be applied automatically to the comment.
```
FredyC: Uh, are you sure that's true about props? Where does that happen? In your third example here you are even converting props to observable despite the `observer`, so that would be fairly contradictory ;)

README.md
```diff
@@ -28,6 +32,121 @@ Project is written in TypeScript and provides type safety out of the box. No Flo
...
+
+`useObserver` keeps the component tree flat as it doesn't introduce a higher order component, but to make props observable, the [`useAsObservableSource`](#useasobservablesourcetstate-t-t) hook should be used.
+
+`<Observer>` is slightly better optimized compared to `useObserver`, but does introduce an additional component in the component tree.
```
FredyC: In case you use it on some enclosed part of the render tree, then yea, it's probably faster, but wrapping a whole component body into it is the same as `useObserver`, isn't?

README.md
```diff
@@ -28,6 +32,121 @@ Project is written in TypeScript and provides type safety out of the box. No Flo
...
+
+```javascript
+import { useLocalStore, useObserver } from "mobx-react-lite"
+
+const P3 = () => {
+    const todo = useLocalStore(() => ({
+        title: "Test",
+        done: true,
+        toggle() {
+            this.done = !this.done
```
FredyC: I have a feeling we should figure out what's wrong with the TypeScript there. I don't feel like recommending `this` in here and then people coming complaining it's not working.

Also honestly, if we forget about TypeScript, isn't it actually better for a user to see `todo.done` in here? I mean these are functional components, one of the selling points is no need for confusing `this`. Why should possibly "scare" people off?

README.md
```diff
@@ -28,6 +32,121 @@ Project is written in TypeScript and provides type safety out of the box. No Flo
...
+    }))
+
+    return useObserver(() => (
+        <h1 onClick={todo.toggle}>
+            {todo.title} {todo.done ? "[DONE]" : "[TODO]"}
+        </h1>
+    ))
+})
+```
+
```
FredyC:
```suggestion
[![Edit MobX in React - useLocalStore](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/llqloyp7z7?fontsize=14)
```

README.md
```diff
@@ -28,6 +32,121 @@ Project is written in TypeScript and provides type safety out of the box. No Flo
...
+When using `useLocalStore`, all properties of the returned object will be made observable automatically, getters will be turned into computed properties, and methods will be bound to the store and apply mobx transactions automatically.
+
+_Note: using `useLocalStore` is mostly beneficial for really complex local state, or to obtain more uniform code base. Note that using a local store might conflict with future React features like concurrent rendering._
```
FredyC: Does it conflict with concurrent mode? I mean it's `useState` underneath so it should be pretty safe. It might leak one observable I suppose, but that shouldn't be such an issue. I wouldn't probably warn about something we are not sure about yet.

README.md
```diff
@@ -28,6 +32,121 @@ Project is written in TypeScript and provides type safety out of the box. No Flo
...
+        <div>
+            Multiplied count: <span>{store.multiplied}</span>
+            <button id="inc" onClick={store.inc}>
+                Increment
+            </button>
+        </div>
+    )
+})
+```
+
```
FredyC:
```suggestion
[![Edit MobX in React - counter example](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/qx99jn2m09?fontsize=14)
```

README.md
```diff
@@ -38,10 +157,10 @@ The rendering in the function will be tracked and automatically re-rendered when
 This can come in handy when needing to pass render function to external components (for example the React Native listview), or if you want to observe only relevant parts of the output for a performance reasons.
 
 ```jsx
-import { Observer, useObservable } from "mobx-react-lite"
+import { Observer, useLocalStore } from "mobx-react-lite"
 
 function ObservePerson(props) {
-    const person = useObservable({ name: "John" })
+    const person = useLocalStore(() => ({ name: "John" }))
```
FredyC: Based on this little code I think people might be fairly confused why not to use `useAsObservableSource`. It would be simpler in this case imo.

README.md
```diff
@@ -212,9 +331,9 @@ function Counter({ multiplier }) {
 
 In the above example, any change to `multiplier` prop will show up in the `observableProps` observable object, and be picked up by the `store`.
 
-Warning: _the return value of `useAsObservableSource` should never be deconstructed! So, don't write: `const {multiplier} = useAsObservableSource({ multiplier })`!_
+Warning: \_the return value of `useAsObservableSource` should never be deconstructed! So, don't write: `const {multiplier} = useAsObservableSource({ multiplier })`!\_useObservable
 
-The value passed to `useAsObservableSource` should always be an object, and is made only shallowly observable.
+The value passeduseObservableuld always be an object, and is made only shallowly observable.
```
FredyC: Wait what it `passeduseObservableuld`? :D

FredyC:
Thanks for doing this!

README.md
```diff
@@ -28,6 +32,121 @@ Project is written in TypeScript and provides type safety out of the box. No Flo
...
+
+```typescript
+import { observer, useAsObservableSource, useLocalStore } from "mobx-react-lite"
+
+interface CounterProps {
+    multiplier: number
+}
+
+export const Counter = observer(function Counter(props: CounterProps) {
+    const observableProps = useAsObservableSource(props)
```
joenoon: Just noticed in the lines about `observer` above and didn't know about this before:

"When using `observer`, the component `props` will be made automatically observable (so that they can be used in reactions and computed values), and `React.memo` will be applied automatically to the comment."

So in this example we wouldn't need `useAsObservableSource` and props would already be IObservable?

FredyC: Yea, I commented on that as well and I don't think it's correct. There is no such magic happening imo.

johnnyreilly: I'm super interested in the answer to this!  Will hopefully get time tomorrow to try and convert over a component to using `mobx-react-lite`

README.md
mweststrate: That is incorrect indeed, was confused with the mobx-react version :see_no_evil:

README.md
mweststrate: I was referring to this: as soon as `useAsObservableSource` is used (or `useLocalStore` with the api change in #142), `useObserver` introduces double re-renderings: https://github.com/mobxjs/mobx-react-lite/issues/94#issuecomment-484085373

README.md
mweststrate: afaik TS inference is normally better with `this`, and hopefully this hints that for examle `useLocalStore(() => new CounterStore())` works, as does extracting that whole function outside the component to avoid re-creating the closure over and over again.

Probably we should add that as tip as well.

README.md
mweststrate: Afaik properly supporting concurrent mode means that it should be possible to rebase local state changes, which is not possible with this solution. Afaik it is not more than honest to warn that this is potientially not compatible with certain future react features, even when I'm not fully vested on their benefits in the general case yet :)

README.md
mweststrate: Me fighting with the scroll wheel :P :computer_mouse:

mweststrate:
Pushed some additional improvements

README.md
FredyC: Yea, so it's in specific scenarios. They way you wrote it feels like it's always faster and people should be favoring that. It's slightly misleading.

README.md
FredyC: Ok, so what are we going to do about it? All examples should use a unified way, otherwise people come asking sometimes it's `this` or `that`. I vote for `that` for reasons I've said above.

README.md
FredyC: Well, `useObserver` also wasn't Concurrent mode ready and we've figured it out. Such a warning feels premature and tells: "you better not to use this, otherwise, you will have to refactor later".

mweststrate:
The readme is a not a marketing brochure, so let's be as honest as possible
:). This is a potential risk and should be advertised as such imho, as it
does diverge from the React's team long term vision for React. While for
the other one I knew there was at least a work around, for this one there
is none as far as I can tell at this point.

FredyC:
> The readme is a not a marketing brochure, so let's be as honest as possible :). This is a potential risk and should be advertised as such imho, as it does diverge from the React's team long term vision for React. While for the other one I knew there was at least a work around, for this one there is none as far as I can tell at this point.

Yea, that's cool, but how useful is to just warn without giving a hint if it can ever be solved. If you want honesty, would you tell to avoid that API in case they need the high performance? Warning alone is fairly useless without giving an alternative solution or a promise of one.

README.md
rvion: I understand that nothing is certain, but as a user, I would much prefer see this in the readme too :)

README.md
rvion: I don't think so, because it is supposed to be mutable.
5 lines below:
```tsx
    <button onClick={() => (person.name = "Mike")}>No! I am Mike</button>
```
so `useAsObservableSource` would not work here.
I think it is fine like this

README.md
```diff
@@ -28,6 +32,123 @@ Project is written in TypeScript and provides type safety out of the box. No Flo
...
+}, 1000)
+```
+
+[![Edit MobX in React - observer pattern](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/4qrv105l47?fontsize=14)
+
+When using `observer`, `React.memo` will be applied automatically to the component, and `useObserver` will be applied under the hood..
+
+`useObserver` keeps the component tree flat as it doesn't introduce a higher order component, but to make props observable, the [`useAsObservableSource`](#useasobservablesourcetstate-t-t) hook should be used.
+
+`<Observer>` is slightly better optimized compared to `useObserver`, as it doesn't cause a double rendering when using `useAsObservableSource`, but does introduce an additional component in the component tree.
```
rvion: Not quite fan of `is slightly better optimized` as it's not quite exact from what I understand.

How about something along the lines of 

- `<Observer>` is basically just `useObserver` wrapped in a new component.
- `<Observer>` should be preffered in conjunction with `useAsObservableSource`, to avoid double rendering by creating component in the component tree.`
- avoid `useObserver` after `useAsObservableSource` as it will cause a `double rendering` issue, etc.

README.md
```diff
@@ -28,6 +32,123 @@ Project is written in TypeScript and provides type safety out of the box. No Flo
...
+        </h1>
+    ))
+})
+```
+
+When using `useLocalStore`, all properties of the returned object will be made observable automatically, getters will be turned into computed properties, and methods will be bound to the store and apply mobx transactions automatically.
+
+_Note: using `useLocalStore` is mostly beneficial for really complex local state, or to obtain more uniform code base. Note that using a local store might conflict with future React features like concurrent rendering._
+
+### More extensive example
```
rvion: small problem here:

> `The following example combines all concepts mentioned so far [...] and useAsObservableProps to make the props observable`

because `useAsObservableProps` has not yet been mentioned in the readme. this is the first time `useAsObservableProps` appears

README.md
```diff
@@ -28,6 +32,123 @@ Project is written in TypeScript and provides type safety out of the box. No Flo
...
+
+1. Make it possible to introduce reactivity to React function components
+2. Make it possible to have local observable state in React function components
+
+This library does _not_ support **class based components**. If you need class based component support, use [mobx-react](https://github.com/mobxjs/mobx-react), which wraps this library.
+
+### Adding reactivity to React components
+
+There are three ways to add reactivity to components: `observer(fn)`, `<Observer>{renderFn}</Observer>` and `useObserver(renderFn)`.
+In most cases, there difference doesn't matter that much, but here is an overview of the three different approaches:
```
jlil: `their` instead of `there`

FredyC: Good spotting. Please use suggestions next time, it's much more helpful.
```suggestion
In most cases, there difference doesn't matter that much, but here is an overview of the their different approaches:
```

FredyC:
Ok, I think I will close this PR since I almost finished converting its contents to the documentation site: https://mobx-react.netlify.com

Eventually, I will trim down the README to point to that site instead.

PR #142, opened 2019-04-30
Allow to pass props to useLocalStore

sheerun:
As I commented in #94 I believe useAsObservableSource is not necessary if its sole purpose is supporting using props in useLocalStore. This PR adds this functionality directly to useLocalStore:

Example:

```es6
function Counter({ multiplier }) {
    const store = useLocalStore(props => ({
        count: 0,
        get multiplied() {
            return props.multiplier * this.count
        },
        inc() {
            this.count += 1
        }
    }), { multiplier })

    return (
        <Observer>
            {() => (
                <div>
                    Multiplied count: {store.multiplied}
                    <button onClick={store.inc}>Increment</button>
                </div>
            )}
        </Observer>
    )
}
```

### DIFF
--- a/src/useLocalStore.ts
+++ b/src/useLocalStore.ts
@@ -11,26 +11,13 @@ function wrapInTransaction(fn: Function) {
     }
 }
 
-export function useLocalStore<T>(initializer: (props?: any) => T, props?: any): T {
-    const [res] = useState(() => (props ? observable(props, {}, { deep: false }) : {}))
-
-    if (typeof props !== "undefined") {
-        if (process.env.NODE_ENV !== "production" && !isPlainObject(props)) {
-            throw new Error("useLocalStore expects an object as second argument")
+export function useLocalStore<T>(initializer: (props?: any) => T, current?: any): T {
+    const local = useState(() => {
+        let props
+        if (isPlainObject(current)) {
+            props = observable(current, {}, { deep: false })
         }
-
-        if (
-            process.env.NODE_ENV !== "production" &&
-            Object.keys(res).length !== Object.keys(props).length
-        ) {
-            throw new Error("the shape of props passed to useLocalStore should be stable")
-        }
-
-        Object.assign(res, props)
-    }
-
-    return useState(() => {
-        const store: any = observable(initializer(res))
+        const store: any = observable(initializer(props))
         if (isPlainObject(store)) {
             Object.keys(store).forEach(key => {
                 const value = store[key]
@@ -39,6 +26,20 @@ export function useLocalStore<T>(initializer: (props?: any) => T, props?: any): 
                 }
             })
         }
-        return store
+        return { store, props }
     })[0]
+
+    if (isPlainObject(current)) {
+        if (
+            process.env.NODE_ENV !== "production" &&
+            Object.keys(local.props).length !== Object.keys(current).length
+        ) {
+            throw new Error("the shape of props passed to useLocalStore should be stable")
+        }
+        Object.assign(local.props, current)
+    } else if (process.env.NODE_ENV !== "production" && typeof current !== "undefined") {
+        throw new Error("useLocalStore expects an object as second argument")
+    }
+
+    return local.store
 }
--- a/src/useLocalStore.ts
+++ b/src/useLocalStore.ts
@@ -1,5 +1,5 @@
 import { observable, transaction } from "mobx"
-import { useState } from "react"
+import { useMemo } from "react"
 import { isPlainObject } from "./utils"
 
 // tslint:disable-next-line: ban-types
@@ -11,9 +11,26 @@ function wrapInTransaction(fn: Function) {
     }
 }
 
-export function useLocalStore<T>(initializer: () => T): T {
-    return useState(() => {
-        const store: any = observable(initializer())
+export function useLocalStore<T>(initializer: (props?: any) => T, props?: any): T {
+    const res = useMemo(() => (props ? observable(props, {}, { deep: false }) : {}), [])
+
+    if (typeof props !== "undefined") {
+        if (process.env.NODE_ENV !== "production" && !isPlainObject(props)) {
+            throw new Error("useLocalStore expects an object as second argument")
+        }
+
+        if (
+            process.env.NODE_ENV !== "production" &&
+            Object.keys(res).length !== Object.keys(props).length
+        ) {
+            throw new Error("the shape of props passed to useLocalStore should be stable")
+        }
+
+        Object.assign(res, props)
+    }
+
+    return useMemo(() => {
+        const store: any = observable(initializer(res))
         if (isPlainObject(store)) {
             Object.keys(store).forEach(key => {
                 const value = store[key]
@@ -23,5 +40,5 @@ export function useLocalStore<T>(initializer: () => T): T {
             })
         }
         return store
-    })[0]
+    }, [])
 }
--- a/src/useLocalStore.ts
+++ b/src/useLocalStore.ts
@@ -1,5 +1,5 @@
 import { observable, transaction } from "mobx"
-import { useMemo } from "react"
+import { useState } from "react"
 import { isPlainObject } from "./utils"
 
 // tslint:disable-next-line: ban-types
@@ -12,7 +12,7 @@ function wrapInTransaction(fn: Function) {
 }
 
 export function useLocalStore<T>(initializer: (props?: any) => T, props?: any): T {
-    const res = useMemo(() => (props ? observable(props, {}, { deep: false }) : {}), [])
+    const [res] = useState(() => (props ? observable(props, {}, { deep: false }) : {}))
 
     if (typeof props !== "undefined") {
         if (process.env.NODE_ENV !== "production" && !isPlainObject(props)) {
@@ -29,7 +29,7 @@ export function useLocalStore<T>(initializer: (props?: any) => T, props?: any): 
         Object.assign(res, props)
     }
 
-    return useMemo(() => {
+    return useState(() => {
         const store: any = observable(initializer(res))
         if (isPlainObject(store)) {
             Object.keys(store).forEach(key => {
@@ -40,5 +40,5 @@ export function useLocalStore<T>(initializer: (props?: any) => T, props?: any): 
             })
         }
         return store
-    }, [])
+    })[0]
 }
--- a/test/useLocalStore.test.tsx
+++ b/test/useLocalStore.test.tsx
@@ -1,9 +1,11 @@
+import mockConsole from "jest-mock-console"
 import * as mobx from "mobx"
 import * as React from "react"
+import { renderHook } from "react-hooks-testing-library"
 import { act, cleanup, fireEvent, render } from "react-testing-library"
 
-import { observer, useLocalStore, useObserver } from "../src"
-import { useEffect } from "react"
+import { Observer, observer, useLocalStore, useObserver } from "../src"
+import { useEffect, useState } from "react"
 import { autorun } from "mobx"
 
 afterEach(cleanup)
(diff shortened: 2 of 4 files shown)


joenoon:
Couple issues just off the top of my head:

1. I think there was previous discussion specifically around not using `useMemo`.  Per the official docs:

> **You may rely on useMemo as a performance optimization, not as a semantic guarantee.** In the future, React may choose to “forget” some previously memoized values and recalculate them on next render, e.g. to free memory for offscreen components. Write your code so that it still works without useMemo — and then add it to optimize performance.

2. Overlooking useMemo and for the sake of conversation lets assume it provided guarantees: The way its used with `[]` might as well just be `useState`.

3. This assumes you only want one `useLocalStore` concerned with props.  The current approach seems to be flexible and separate the concerns: step 1) get observable props.  step 2) use it as an observable in any way you wish (one or many other observables, or just pass it along to another component).

sheerun:
1 2 - OK I thought it doesn't matter. I've changed useMemo to useState

3 - It's your judgement if you want to merge this PR. All I'm saying it **is** possible and this approach doesn't require second import for very common use case

xaviergonz:
Also, I think if you want to use mobx effects (reaction, etc) that use both localState and props then you need to add the same to those, thus having to convert props to an observable object two times.

FredyC:
I think I kinda do like it. It's closer to React API when you specify non-observable dependencies and yet it won't throw away the previous state. Feels like it could be actually viable indeed solution.

Btw, why not to use `useAsObservableSource` inside this hook? No need to duplicate stuff. I know you are trying to remove that second hook, but...

> Also, I think if you want to use mobx effects (reaction, etc) that use both localState and props then you need to add the same to those, thus having to convert props to an observable object two times.

For this reason, both hooks should be kept so you have a choice if you need to use observable props in other ways and not just stores.

Although, you could also expose those props on the `store` itself if you need to access it from the reaction. So the `useLocalStore` would become more of the central point of all things observable and everything would be coming from it. Interesting :)

sheerun:
> No need to duplicate stuff. I know you are trying to remove that second hook, but...

Indeed it was my intention but if it's meant to stay I can use it directly.\

I think also typescript type need refinement as currently it's `all` and I'm pretty sure typescript can allow derivation of types so props received in `useLocalStore` have the same type as props passed to it. I tried following but I've got lot of errors and I'm no typescript expert:

```es6
useLocalStore<T, P>(initializer: (props?: P) => T, props?: P): T 
```

sheerun:
Actually it turns out I can't use `useAsObservableSource` from useLocalStore as error messages don't match. Also if I use it directly then observable is called second time even if props are not passed which breaks existing tests of useLocalStore. Probably it's better to leave it as it is.

sheerun:
(and I can't use useAsObservableSource inside `if` expression because react hooks guide explicitly says I shoudn't do it: all hook calls should be at "top level" of component's code)

mweststrate:
I like it! The first iteration of `useLocalStore` was similar, except the arguments were swapped whcih looked, and it didn't provide a good answer to where to leave reactions. 
This proposal makes the common case much simpler. 

I agree with @FredyC it should use `useAsObservableSource` under the hood, reusing it, but with adjusted error messages is easy enough when factoring out a common function with the implementation.

src/useLocalStore.ts
```diff
@@ -11,9 +11,26 @@ function wrapInTransaction(fn: Function) {
     }
 }
 
-export function useLocalStore<T>(initializer: () => T): T {
+export function useLocalStore<T>(initializer: (props?: any) => T, props?: any): T {
+    const [res] = useState(() => (props ? observable(props, {}, { deep: false }) : {}))
+
+    if (typeof props !== "undefined") {
```
mweststrate: Let's reuse `useAsObservableSource`

src/useLocalStore.ts
```diff
@@ -11,9 +11,26 @@ function wrapInTransaction(fn: Function) {
...
+        ) {
+            throw new Error("the shape of props passed to useLocalStore should be stable")
+        }
+
+        Object.assign(res, props)
+    }
+
     return useState(() => {
-        const store: any = observable(initializer())
+        const store: any = observable(initializer(res))
```
mweststrate: `res` -> `observableProps` (or something else a bit clearer)?

test/useLocalStore.test.tsx
```diff
@@ -236,4 +238,210 @@ describe("is used to keep observable within component body", () => {
...
+
+    describe("with props", () => {
+        it("and useObserver", () => {
+            let counterRender = 0
+            let observerRender = 0
+
+            function Counter({ multiplier }: { multiplier: number }) {
+                counterRender++
+
+                const store = useLocalStore(props => ({
```
mweststrate: When updating the Readme, we should make it abundantly clear that `props` isn't supposed to be destructured

mweststrate: Before updating the readme, it is probably the best to wait until #140 is merged to avoid potential merge issues

src/useLocalStore.ts
sheerun: As mentioned in https://github.com/mobxjs/mobx-react-lite/pull/142#issuecomment-488320376 it can't really be done

mweststrate: I think the common stuff can still be extracted, just parameterize the hook name in the error message, and make sure the call from this hook can deal with the absence of the thing?

sheerun: This seems like overkill to avoid copying 2 lines of code (rest is just development-only). Parametrizing messages will also make typescript type of `useAsObservableSource` very ugly and I'll need to check whether `current` is defined each time before I do `Object.assign` which currently is not necessary as `useAsObservableSource` in non-development mode just assumes that current is an object.

sheerun: I'll commit in short while code that reduces number of hook calls inside `useLocalStore` to 1 and still supports passing custom props dependencies

mweststrate:
No need to touch the public api, the common core implementation can be
extracted and wrapped. But if you like it better this way, fair enough.

sheerun:
I've pushed updated code

sheerun:
@mweststrate I'm thinking if it's not better to make second argument update store itself instead.. So the usage would be as follows:

```es6
function Counter({ multiplier }) {
    const store = useLocalStore(() => ({
        count: 0,
        get multiplied() {
            return this.multiplier * this.count
        },
        inc() {
            this.count += 1
        }
    }), { multiplier })

    return useObserver(() => 
        <div>
            Multiplied count: {store.multiplied}
            <button onClick={store.inc}>Increment</button>
        </div>
    )
}
```

So this way useLocalStore would have following signature:

```typescript
useLocalStore(initializer: () => any, updates?: any)
```

sheerun:
(this way we avoid "please don't destructure this" problem, because this way `this` is forced)

mweststrate:
Interesting though! Not sure what to think about it though as of yet.
Typing will be an issue at least it seems :)

FredyC:
> I'm thinking if it's not better to make second argument update store itself instead..

That sounds too magical to me. How would you solve name clashes? On the contrary, it should be within your power to explicitly specify which properties are to be exposed as observable, which one is used for a derivation and some are for actions only.

As I said above, I kinda like that you don't really need `useAsObservableSource` because you can expose anything you like with `useLocalStore`. Although the `useAsObservableSource` is definitely easier for simple cases (eg. running reactions based on props), so I would keep it anyway.

src/useLocalStore.ts
```diff
@@ -11,26 +11,13 @@ function wrapInTransaction(fn: Function) {
...
 }
 
-export function useLocalStore<T>(initializer: (props?: any) => T, props?: any): T {
-    const [res] = useState(() => (props ? observable(props, {}, { deep: false }) : {}))
-
-    if (typeof props !== "undefined") {
-        if (process.env.NODE_ENV !== "production" && !isPlainObject(props)) {
-            throw new Error("useLocalStore expects an object as second argument")
+export function useLocalStore<T>(initializer: (props?: any) => T, current?: any): T {
+    const local = useState(() => {
```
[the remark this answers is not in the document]
sheerun: I think I've just preserved an convention that was here before: https://github.com/mobxjs/mobx-react-lite/blob/abdf573a343ce43cb5cbf7318202579f80990f35/src/useLocalStore.ts#L26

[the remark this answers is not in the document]
FredyC: Yea well, it kinda did make sense there because it was returning that value directly, but when you working with I think it makes more sense to destructure.

sheerun:
I have the same doubts about updating store itself so I think current form or this PR is OK.

The only remaining issue I have about mobx-react-lite (if I'm understanding it correctly) is that wrapping just JSX with with `useObserver` or `<Observable>` makes it unresponsive to changes in props themselves (ones that are not wrapped with `useObservableSource`). This makes me think that `observer` is be most reasonable function from this package because it wraps whole component with `useObserver`, making any access of props reactive by mobx (missing only "local state" use case).

This leads me to final conclusion that maybe `useLocalStore`, `useObserver` and `<Observable>` is also not necessary and this package could be only export `observer` function that accepts "local store" initializer second argument:

```es6
export default observer(
  // mobx-react-lite passes observable object to this function
  function Counter ({ multiplied, inc }) => (
    <div>
      multiplied count: {multiplied}
      <button onclick={inc}>increment</button>
    </div>
  ),
  // maps component's props to observable object that is passed to the function above
  ({ multiplier }) => ({
    count: 0,
    get multiplied() {
      return multiplier * this.count
    },
    inc() {
      this.count += 1
    }
  })
)
```

Above soultion also has no "don't destructure it" problem and in my opinion encourages good react practices (cleanly separating "pure" component from "state")

I know this PR is close to completion, but I'm thinking to back out of this in lieu of above..

sheerun:
Uh sorry, props passed to second argument actually have "don't destructure this" problem, so:

```es6
export default observer(
  // mobx-react-lite passes observable object to this function
  function Counter ({ multiplied, inc }) => (
    <div>
      multiplied count: {multiplied}
      <button onclick={inc}>increment</button>
    </div>
  ),
  // mobx-react-lite converts component's props to observable props and passes it this:
  props => ({
    count: 0,
    get multiplied() {
      return props.multiplier * this.count
    },
    inc() {
      this.count += 1
    }
  })
)
```

FredyC:
We already had a long discussion on which observer is better (mainly in #94), but for now, we are gonna keep all of them. If you want, you can join a little project of gathering pros & cons about this... https://www.kialo.com/differences-with-mobx-observer-pattern-in-react-28353

> This makes me think that `observer` is be most reasonable function from this package because it wraps whole component with `useObserver`, making any access of props reactive by mobx (missing only "local state" use case).

How does `observer` make props observable? I think you got confused there, there is no such magic. The whole component will re-render with new props the same way as any other component, that's unrelated to MobX.

sheerun:
Well, in current form indeed it won't. It could be arranged if you changed behavior `observer` so component wrapped with it would receive props wrapped with `observable` instead of raw props.

Sorry I'm out of loop on #94, it's very long conversation to read, not mention to understand.

mweststrate:
I think it wouldn't compose as nicely as useLocalStore currently is. I
think this PR is in great shape atm.

 Didn't really get what was meant with the changes in props don't propagate
as described before. (But currently on phone, should probably read another
time)

Op do 2 mei 2019 22:36 schreef Adam Stankiewicz <notifications@github.com>:

mweststrate:
> ) is that wrapping just JSX with with useObserver or <Observable> makes
it unresponsive to changes in props themselves (ones that are not wrapped
with useObservableSource).

That shouldn't be true, a new closure is created in those case, causing
useObserver to render. Can you provide an example of a case that doesn't
work?

Op do 2 mei 2019 23:16 schreef Michel Weststrate <mweststrate@gmail.com>:

sheerun:
@mweststrate @FredyC Here's working implementation of `observer` I imagined above on top of current mobx-react-lite:

https://codesandbox.io/s/23x87224xr

sheerun:
@mweststrate OK I actually tried to reproduce "unresponsive" and I was not able to. Neverthless I still think the `observer` API I presented above is everything mobx-react-lite needs for most use cases. I guess other methods could stay as "plumbing commands".

Please see how `useLocalStore` requires you to change JSX so it references `store.___` variables instead. In `observable` solution component code is exactly the same.

e.g. if there is stateless "controlled" component:

```es6
function Counter(({ multiplier, multiplied, inc }) => (
  <div>
    Multiplier: {multiplier}
    Multiplied count: {multiplied}
    <button onClick={inc}>increment counter</button>
  </div>
));
```

in case of useLocalStore can be converted to stateful component with something like:


```es6
const WrappedComponent = ({ multiplier }) {
  const observableProps = useAsObservableSource({ multiplier });

  const store = useLocalStore(() => ({
    count: 0,
    get multiplied() {
      return observableProps.multiplier * this.count;
    },
    inc() {
      this.count += 1;
    }
  }));

  return <Counter multiplier={multiplier} {...store} />
}
```

while with `observer` it can be:

```es6
const WrappedComponent = observer(Component, props => ({
  count: 0,
  get multiplied() {
    return props.multiplier * this.count;
  },
  inc() {
    this.count += 1;
  }
}))
```

I hope you notice the appeal

FredyC:
> OK I actually tried to reproduce "unresponsive" and I was not able to. Neverthless I still think the `observer` API I presented above is everything mobx-react-lite needs for most use cases. I guess other methods could stay as "plumbing commands".

I am using `useObserver` quite heavily, especially in other hooks which I want to ensure will be observable on its own. You can never do that with `observer`. Please if you don't want to contribute to the pros&cons project then stop discussing this, this is a bad place for such a discussion.

The variant of `observer` you are suggesting is a better fit for user-land code. It's too convoluted and doesn't compose that well.

sheerun:
👍

FredyC:
So, we shall figure out how to proceed here. I would be inclined toward removing the `useAsObservableSource`, but I was thinking how would someone approach the following scenario with just the `useLocalStore` hook and it's not that pretty.

```jsx
const WrappedComponent = ({ multiplier }) {
  const globalStore = useStore()
  const store = useLocalStore(() => ({ multiplier }), { multiplier })
 
  React.useEffect(() => autorun(() => {
    alert(store.multiplier * globalStore.count)
  }), [])

  return <whatever />
}
```

Without that hook, it's a lot of duplicate statements and confusing constructs. Keeping the hook would mean I can do the following which definitely reads better. _Remaining code is the same, so leaving out for brevity_

```jsx
  const store = useAsObservableSource({ multiplier })
```

So yea, let's keep it and try to reuse at least some code from within the `useLocalStore`.

@sheerun Are you able to finish it or should I take over?

sheerun:
@FredyC I've resolved suggestion about destrucuring state, but I'd rather leave README updates (probably you'd need to fix them anyway) and typescript types to you (not competent enough).

FredyC:
@sheerun Don't worry about README, everything from there is being moved to new site https://mobx-react.netlify.com. Feel free to submit a PR there with changes regarding this.

Ok, I will wrap it up here sometimes in this week hopefully. Thanks for the work.

src/useLocalStore.ts
```diff
@@ -11,9 +11,13 @@ function wrapInTransaction(fn: Function) {
     }
 }
 
-export function useLocalStore<T>(initializer: () => T): T {
-    return useState(() => {
-        const store: any = observable(initializer())
+export function useLocalStore<T>(initializer: (props?: any) => T, current?: any): T {
```
mweststrate: I think this type suffices:
`export function useLocalStore<T, P = any>(initializer: (props?: P) => T, current?: P): T {`

FredyC:
Closing in favor of #159

FredyC:
The change is now documented here: https://mobx-react.netlify.com/state-local#non-observable-dependencies

Issue #143, opened 2019-05-01
regression: observable.map no longer updates observers in 1.3.0


pdeva:
sample code 
```js
       @observable
       filterValues = observable.map({}, {deep: false, name: "filterValues"});

        @action
	setFilterValue(filterName: string, values: string[])
	{
		const strings = values.filter(v => v != null);
		this.filterValues.set(filterName, strings);
	}
```

this works properly and updates listening components in 1.2.0 but does not update them in 1.3.0

FredyC:
Um, I don't see any mobx-react-lite related code in here. Please provide codesandbox with the actual issue.

pdeva:
@FredyC the reason its mobx-react-lite related is because just changing mobx-react-lite from 1.2.0 to 1.3.0 causes the issue to surface.

i removed the decorator from the observable.map field, and yes it does work normally in 1.2.0 but the issue remains in 1.3.0.

not sure how to create a sandbox for this, since our codebase is so huge and complicated. is there any info, piece of code i can provide that can help get to the bottom of this?

FredyC:
Sorry, cannot help it until you show the code that's related to the package.

pdeva:
here is some code that uses the mobx-react-lite `observer` to access the `filterValues` field shown in the OP:


```ts
const EntityFilter = observer(({column}: IFilterValueFetcherProps) => {

        //Another observable accessed here : 'filters'
	const defs = clusterFilterStore.filters.filter(f => f.column == column);

	let initialChecked = true, initialSelected: string[] = [];
	if (defs.length > 0)
	{
		initialChecked = defs[0].include;
		initialSelected = defs[0].items.slice(); //slice cause its a proxy
	}

	const [isIncludedChecked, setIsIncludedChecked] = useState(initialChecked);
	const [selectedItems, setSelectedItems] = useState<string[]>(initialSelected)

        //USES filterValues here
	const list = cubeStateStore.filterValues.get(column);
	const vals: string[] = list ? list : [];
	const filteredItems = vals.filter(it => !selectedItems.includes(it));

	return (
		<>
			{
				(cubeStateStore.cluster && cubeStateStore.entity) &&
				<FilterValueFetcher key={column} column={column}/>
			}
			<Radio.Group value={isIncludedChecked} onChange={onRadioGroupChange}>
				<Radio value={true}>Include</Radio>
				<Radio value={false}>Exclude</Radio>
			</Radio.Group>
[...]

PR #145, opened 2019-05-01
Revert "Add test for proper reaction disposal in StrictMode"

FredyC:
Reverts mobxjs/mobx-react-lite#119

This was surpassed by #121 and should not have been released. It's actually possible it is a culprit of #143 even though current tests are passing. 

@RoystonS I think you said you had some problems with this change, right? So it's probably better to get rid of it altogether.


RoystonS:
#119 by itself was problematic, but #121 was based on #119

RoystonS:
(It would be good to have some failing tests for #143)

src/useObserver.ts
```diff
@@ -25,25 +25,17 @@ export function useObserver<T>(
...
-            if (committed.current) {
-                forceUpdate()
-            }
+            forceUpdate()
         })
     }
 
     useDebugValue(reaction, printDebugValue)
 
-    useEffect(() => {
```
meyer: Changing this `useEffect` hook to `useLayoutEffect` fixes the bug for me.

FredyC: We certainly don't want to `useLayoutEffect`, that's bad for karma and stuff :) Besides, you said this PR alone fixed it, so I am not sure why would modify it even further. Please do provide an example, this is really hard to understand what kind of issue it actually is.

meyer: ha, yeah, makes sense. Out of curiosity, what’s the reason that `committed.current` is set _asynchronously_ in the `useEffect` hook rather than _synchronously_ in the body of the `useObserver` hook?

meyer: Moving `committed.current = true;` out of the `useEffect` hook also fixes the bug I was encountering.

FredyC: Don't worry about this. It was really just an early experiment. See #121 for a proper solution and some details.

meyer: rad, thanks 🎉

PR #156, opened 2019-05-12
Trim down README and direct to docz site

FredyC:
I will keep this issue open for like 2 weeks so there is enough time to review the content of the site: https://mobx-react.js.org

Let's keep it under the lid till it can be considered for quality docs for a public release.

Please spread the word to people you know can help with reviewing.

@mweststrate @xaviergonz @johnnyreilly @RoystonS @joenoon @urugator @luisherranz @mattruby

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,228 +8,55 @@ This is a next iteration of [mobx-react](https://github.com/mobxjs/mobx-react) c
 
 **You need React version 16.8.0 and above**
 
-Class based components **are not supported** except using `<Observer>` directly in its `render` method. If you want to transition existing projects from classes to hooks (as most of us do), you can use this package alongside the [mobx-react](https://github.com/mobxjs/mobx-react) just fine. The only conflict point is about the `observer` HOC. Subscribe [to this issue](https://github.com/mobxjs/mobx-react/issues/640) for a proper migration guide.
+Class based components **are not supported** except using `<Observer>` directly in class `render` method. If you want to transition existing projects from classes to hooks (as most of us do), you can use this package alongside the [mobx-react](https://github.com/mobxjs/mobx-react) just fine. The only conflict point is about the `observer` HOC. Subscribe [to this issue](https://github.com/mobxjs/mobx-react/issues/640) for a proper migration guide.
 
 [![NPM](https://nodei.co/npm/mobx-react-lite.png)](https://www.npmjs.com/package/mobx-react-lite)
 
 Project is written in TypeScript and provides type safety out of the box. No Flow Type support is planned at this moment, but feel free to contribute.
 
--   [API documentation](#api-documentation)
-    -   [`<Observer/>`](#observer)
-    -   [`observer<P>(baseComponent: FunctionComponent<P>, options?: IObserverOptions): FunctionComponent<P>`](#observerpbasecomponent-functioncomponentp-options-iobserveroptions-functioncomponentp)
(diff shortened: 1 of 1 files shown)


rvion:
few remarks :

- https://mobx-react.netlify.com/state-how#creating-a-state : `useLocalState` instead of `useLocalStore` 
image

- poor contrast: 
image or 
image or 
image. not readable on bad screens.


- examples are cut for no reasons (need to scroll to see the full code) 
image  => surprise, there is 42 lines and not 15, and no content further down the page.

- search is broken & deficient:
image, or 
image

- also, several pages  =>  `ctrl-f` not working to easilly find references
- no full index page (level 3 titles are only visible when in level 1 pages)

I'm overall not quite convinced this is a good idea. I agree dedicated doc sites make sense when there is specific code /doc browsing enriched experience. Here, I'm wondering if github readme is just enough.
this being said, I'm absolutely no documentation expert!

thanks for all the hard work anyway.

FredyC:
Thanks for the feedback.

Well, styling is something I hope I will get help with. I am no designer, I just grabbed the default dark theme of Docz and it looks good to me. It's possible to tweak every color general styling of everything, it's a very customizable platform. The examples display can be surely tweaked as well.

The search looks only in titles of documents sadly. Since it's statically generated site, it's much harder to do a proper full-text. It's not impossible for sure, but not the main focus right now. I am trying to build it more like a user guide than an API reference. Perhaps I will keep API reference in README with links to the site for a more explanation.

The README approach is problematic for really large documents. Sure, you can split it into multiple files, but searching doesn't work that well. And you cannot include interactive examples as I did there.

Besides, that site is lib agnostic, it's not for `mobx-react-lite` only, but it essentially covers the `mobx-react` and possibly some utils that will appear in the future. It wouldn't make any sense to keep docs in README of a single repo.

ThomasRoest:
Looks good! Here are some things I've noticed ( keep in mind I'm new to mobx )

- The intro example could use a bit more explanation on how this implementation relates to the core concepts. There's no mention of observables etc. 
- The term useLocalStore is a bit confusing. Because it's likely that most people ending up on this page are looking for a global state management solution. (or not?)  And from this site only, it's not entirely clear that this is the right choice. This sentence / example als adds to the confusion
`In short, we don't recommended global state.` in https://mobx-react.js.org/state-how

FredyC:
Thanks for the feedback.

> observables, actions, decorators?

Everything still works as you are used it, we are just providing some utilities to make it easier to React world.

> The term useLocalStore is a bit confusing.

Do you have a better name? The point is it's a state created locally within a component. If you stash it inside the Context or pass around in props, it's up to you. You can still create the state the way you are used to (classes, decorators, etc...).

> most people ending up on this page are looking for a global state management solution

Sure, that works, although it's kinda misleading to call it a "global". Nothing is really global in React world. That paragraph in docs is referring to state stored in a purely global variable. It could probably use better wording.

Ciantic:
Is there a reason why you use todoRef instead of controlled input? Would this work as well:

```typescript 
let newTodo = useLocalStore(() => { value: "" });
```

then 

```typescript
<input value={newTodo.value} />
```

Or is there a shorter way to use observable for a single value like that?

FredyC:
> It could be just a matter of taste

Exactly :) In this case, I feel like it helps to focus on the essence of example. You don't need to scroll back and forth to realize how it works. Both approaches would definitely work.

PR #159, opened 2019-05-16
Allow to pass source to useLocalStore 

FredyC:
A polished version of #142 made by @sheerun

A most notable change is the reuse of `useAsObservableSource` and improved Typescript. Also instead of calling it `props` it's named more universally `source`.

### DIFF
--- a/src/useAsObservableSource.ts
+++ b/src/useAsObservableSource.ts
@@ -1,20 +1,33 @@
 import { observable } from "mobx"
-import { useState } from "react"
+import React from "react"
 
 import { isPlainObject } from "./utils"
 
-export function useAsObservableSource<T>(current: T): T {
+export function useAsObservableSourceInternal<TSource>(
+    current: TSource,
+    usedByLocalStore: boolean
+): TSource {
+    const culprit = usedByLocalStore ? "useLocalStore" : "useAsObservableSource"
+    if (usedByLocalStore && current === undefined) {
+        return undefined as any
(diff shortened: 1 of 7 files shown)


FredyC:
Alright, since nobody has objections, I am going proceed with this and deal with possible issues later :)

sheerun:
I've already written something like this in #142, but it would be nice if `mobx-react` had similar helper for both classes and "stateless" components as decorator. I have many components that I cannot afford to rewrite to functional components and hooks, but I like "local store" / "properties manager" idea (i.e. object that maps props to observable props for stateless component so it's easier to extract than setState and easier to test because component doesn't use `this.state`). Something like this:

```es6
@store(props => ({
  count: 0,
  get multiplied() {
    return props.multiplier * this.count
  },
  inc() {
    this.count += 1
  }
})
class Counter extends React.PureComponent {
  render() {
    const { count, multiplied, inc } = this.props

    return <div>
      count: {count}
      multiplied count: {multiplied}
      <button onclick={inc}>increment</button>
    </div>
  }
}
```

As for this PR: 👍

FredyC:
Well, as much as I would agree, this definitely does not fit into this package. We won't be supporting classes here for sure. Please open an issue at `mobx-react` repo and perhaps some version of it can get to upcoming V6.

For what it's worth, I think it's fairly simple to make something like that in user-land since the HOC can `useLocalStore` just fine.

PR #160, opened 2019-05-16
useObserver reaction dispose check

FredyC:
I would have thought that MobX would not be annoyed by double disposal, but I just ran into it during Cypress tests. I guess things are much faster there.

image

@mweststrate @RoystonS

### DIFF
--- a/src/useObserver.ts
+++ b/src/useObserver.ts
@@ -1,5 +1,6 @@
 import { Reaction } from "mobx"
 import { useDebugValue, useRef } from "react"
+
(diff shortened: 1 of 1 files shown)


mweststrate:
I think this should be fixed in mobx. I think it just got enabled once to
ensure certain internal consistency, and never got disabled.

mweststrate:
(it is fine to fix here as well, at least  for compatibility reasons)

mweststrate:
The above error is fixed in mobx 5.10.0

Issue #170, opened 2019-05-31
runInAction is necessary in LocalStore's (action) methods


asfernandes:
I'm following https://mobx-react.netlify.com/state-local and adapting a existing mobx-react app to use mobx-react-lite.

In this page it says:

> All properties of the returned object will be made observable automatically, getters will be turned into computed properties, and methods will be bound to the store and apply mobx transactions automatically. If new class instances are returned from the initializer, they will be kept as is.

My app uses mobx in strict mode.

I created a simple LocalStore with:

```ts
	const inputStore = useLocalStore(() => ({
		value: 'test',

		setValue(value: string) {
			this.value = value;
		}
	}));
```

But my `setValue` need to use `runInAction`, otherwise I have a strict-mode error.

I understand from the documentation that `runInAction` would not be necessary, as it says it run methods in mobx transactions.

FredyC:
@mweststrate did this part and I don't yet fully understand it yet, so best if he can bring some light here.

Relevant PR: #139

asfernandes:
If I understand the PR correctly, the change was intentional and explicity action would then be necessary?

Is it recommended or not to use mobx in strict-mode currently?

mweststrate:
Actions in Mobx do 3 things:

1. wrap updates in transactions
2. allow modifications (if in strict mode)
3. disable tracking (so that they can be called from e.g. autorun without side effect

Because we cannot determine whether a method in a local store is used as an 'action' or as a 'view', we don't automatically convert them to actions (as 3. makes it unusable as view). However, since we can achieve 1. without any downsides, that is the only thing that is applied to all store methods automatically (+ binding). 

So if you need to use them in strict mode, you will need to be explicit about actions, `inc: action(function() { this.counter ++ })` (note that you can't combine arrow functions with `this`!). 
Alternatively, or actually the best practice; put `action` around the event handlers  that invoke the store methods instead.

asfernandes:
@mweststrate I didn't read any of mobx-react-lite code, but wouldn't it be possible to have the "wrapped" methods in a level above the original ones (using proxy or another way)? In this way, internally called methods (when get calls a method) will call the original method, but external calls will call the wrapped action.

mweststrate:
No, it could also be an externally called view, the get was just an example
to demonstrate the problem.

Op vr 31 mei 2019 21:24 schreef Adriano dos Santos Fernandes <
notifications@github.com>:

FredyC:
[...]

PR #172, opened 2019-06-12
Add support for Mobx enforceActions configuration

johnrichter:
Related to #170. Wraps observable field assignments with `runInAction()` to work in `enforceActions: 'always'` mode (and `never`/`observed`).


test/useAsObservableSource.test.tsx
```diff
@@ -356,3 +356,30 @@ it.skip("checks for stable shape of object being passed in", async () => {
     )
     restore()
 })
+
+describe("enforcing actions", () => {
```
FredyC: I think it's unnecessary to test `useAsObservableSource` in this case since it's used `useLocalStore` internally. It shouldn't break coverage.

test/useLocalStore.test.tsx
```diff
@@ -455,3 +455,69 @@ describe("is used to keep observable within component body", () => {
         restore()
     })
 })
+
+describe("enforcing actions", () => {
+    it("'never' should work", () => {
+        mobx.configure({ enforceActions: 'never' })
+        const restore = mockConsole() // to ignore React showing caught errors
```
FredyC: Are there some actual errors printed? Since all three cases, you expect an error to be undefined, it feels like you are trying to hide something nefarious here :)

test/useLocalStore.test.tsx
```diff
@@ -455,3 +455,69 @@ describe("is used to keep observable within component body", () => {
...
+                    get mul() {
+                        return this.count * props.multiplier
+                    }
+                }),
+                { multiplier: 2 }
+            )
+        })
+        expect(result.error).toBeUndefined();
+        restore()
+        mobx.configure({})
```
FredyC: I think this isn't doing anything actually since no value for `enforceActions` is being passed.

https://github.com/mobxjs/mobx/blob/a5cb65926621266716f8d11d716bf337e60dc3c1/src/api/configure.ts#L25

Probably the best to use `afterEach` and reset it specifically to `never` as default.

test/useLocalStore.test.tsx
johnrichter: To be honest I just looked for the simplest existing tests, copied them, and modified them slightly. I'm not sure about React showing errors comment. Looking the test I coped from again, I can see now that it doesn't make sense to use the mock console as we're not testing errors that are exposed to the user. Given the way `renderHooks` returns results I'm not sure how to clarify the intention more than just the undefined check. Here is what I'm thinking:

```ts
it("'always' should work", () => {
    configure({ enforceActions: 'always' })
    const { result } = renderHook(() => useAsObservableSource({ hello: 'world' }))
    expect(result.error).not.toBeDefined();
})
```

Screen Shot 2019-06-13 at 10 15 22

test/useAsObservableSource.test.tsx
johnrichter: I found that both `useAsObservableSource` and`useLocalStore` modify observables directly in different ways. Since someone could call `useAsObservableSource` directly, I think it makes sense to test both hooks independently. It is also possible that `useAsObservableSource` could be removed from `useLocalStore` in the future (however unlikely) reducing coverage and introducing a regression.

I took another look to see if I could optimize the `runInAction` call out (internal to `useAsObservableSourceInternal`) if `usedByLocalStore` was true, but since `useLocalStore` needs to call `useAsObservableSource` at the top level I don't think I can since I'd need to wrap it in `runInAction` to make it work ☹️

test/useLocalStore.test.tsx
johnrichter: Good call. I should have looked at the source a bit more closely :)

test/useLocalStore.test.tsx
johnrichter: The screenshot is with the new `runInAction` code temporarily removed so I could show the error.

FredyC: Yes, I think that's a much better way like that.

johnrichter:
@FredyC I added a bit of code to each test to ensure that updates to the observables also succeed in each `enforceActions` mode. This is in reference to: https://github.com/mobxjs/mobx-react-lite/issues/170#issuecomment-502165994

iaroslavshvets:
Hi! I would appreciate if you could update this branch to current master, so it could be merged. It totally prevents from using useLocalStore. I forked this branch and updated to current master without any issues. I can create another updated pull request, but I see no point, as you did the job and the only thing missing - it's outdated.

johnrichter:
Thanks for the update and merge @FredyC!

PR #186, opened 2019-07-14
Change links from netlify url to mobx-react.js.org

moismat:


FredyC:
Thanks, I thought I got it everywhere :)

PR #199, opened 2019-08-09
Upgrade React to 16.9

FredyC:
Something has changed in React that's causing one test to fail. I've spent quite some time to figure it out, but I have no idea. I've extracted the failing portion to sandbox and hoping for an extra pair of eyes to see to the problem...

https://codesandbox.io/s/little-fire-2k5ow?expanddevtools=1&fontsize=14&module=%2Fsrc%2Ferror.test.tsx&previewwindow=tests

I tried changing the following line to `createElement(component, props)` but it breaks a whole lot of other tests simply because `useObserver` cannot see into a nested component.

https://github.com/mobxjs/mobx-react-lite/blob/9e8061861310c823e788ab7e1b22c6f635ba7d52/test/observer.test.tsx#L22

image

cc @mweststrate @urugator @xaviergonz 

Fixes #195 #196 #198


mweststrate:
Sorry, I've been really swamped by our migration lately, let me know if
you're still stuck and I'll try to find some time next week

FredyC:
Yep, still stuck, no idea what's the problem here.

FredyC:
Ok, I had a fresh new look at this and something really interesting/strange is happening here. Feels like some partial concurrent mode got into 16.9 or whatnot.

What happens is when an [error is thrown](https://github.com/mobxjs/mobx-react-lite/blob/cc2b30fd6fa739b6b8ba8d0263b62bc4d7952dc3/test/observer.test.tsx#L449) from the observed component, it is caught for reaction disposal and then rethrown again which is fine.

https://github.com/mobxjs/mobx-react-lite/blob/cc2b30fd6fa739b6b8ba8d0263b62bc4d7952dc3/src/useObserver.ts#L60

However, React does another re-render afterward where Reaction is already disposed so it skips rendering and ultimately it returns `undefined` as the initial value of `rendering` variable.

I will try to isolate this to pure React and possibly report an issue with them.

FredyC:
Aha! It seems when I use the code for `useObserver` from the `next` branch, it actually passes a test in this branch. It's indeed concurrent more related.

So I am thinking we will expedite the release of 2.0 as per https://github.com/mobxjs/mobx-react-lite/pull/121#issuecomment-486806528 there might be potentially breaking changes. The 1.x will stick to 16.8.

Issue #73, opened 2019-03-04
Using const MyComponent = observer((props) => to create observer component makes unnamed component <Memo(wrappedComponent)>  in React DevTools


uffou:
I know useObserver is the preferred approach but still, sometimes observer is needed, is this a known issue are there plans to get the proper name instead of this?
![image](https://user-images.githubusercontent.com/2581788/53761810-18fe9280-3ecf-11e9-88c0-05adfbb9963f.png)
![image](https://user-images.githubusercontent.com/2581788/53761992-8dd1cc80-3ecf-11e9-8ca1-4e34273c8fa6.png)

FredyC:
What version are you using? @mweststrate has been fiddling around this recently #70. The 1.0.1 version might be better in theory.

MatejBransky:
I have the similar problem. I'm using `mobx-react-lite` version `1.1.0`.

React DevTools:
![image](https://user-images.githubusercontent.com/20944757/53872721-1b600a00-3fff-11e9-94bb-ee8d14030938.png)
Source code:
```jsx
import React, { useContext } from 'react';
import { observer } from 'mobx-react-lite';

import TimerContext from './TimerStore';

const TimerView = observer(() => {
  const timerStore = useContext(TimerContext);

  return (
    <button onClick={timerStore.reset}>
      Seconds passed: {timerStore.timer}
    </button>
  );
});
```

FredyC:
Yea, the last approach is pretty much what I wanted to suggest. I don't generally like default exports though, so it's an annoying tradeoff.

I think this could be solved by writing a simple macro for `babel-plugin-macros`, but I don't have experience with AST stuff. Volunteers? :)

uffou:
@MatejBransky @FredyC thanks, I will try with export default

mweststrate:
These two approaches will both work to create non-default named exports:

```javascript
let MyComponent1_ = () => "hello1!"; // N.B. 'MyComponent1_' is what shows up in devtools
export const MyComponent1 = observer(MyComponent1_);

export const MyComponent2 = observer(() => "hello2!");
MyComponent2.displayName = "MyComponent2";
```

Demo: https://codesandbox.io/s/w038r5q68w

mweststrate:
This works as well: 

```javascript
let MyComponent4 = () => "hello4!";
MyComponent4 = observer(MyComponent4);
export { MyComponent4 };
```

FredyC:
[...]

PR #203, opened 2019-08-12
Add observer macro

bernhardc:
This solves #73.

The macro transforms anonymous wrapped component declarations like
```jsx
const MyComponent = observer(() => { return <div/> })
```
to
```jsx
const MyComponent = observer(function MyComponent() { return <div/> })
```

To use it, the import statement, e.g.
```
import { observer } from "mobx-react-lite";
```
has to be replaced with
```
import { observer } from "mobx-react-lite/dist/macro";
```
The macro touches only the `observer()` function and passes through all others, so users could just always use the `/macro`imports, like
```
import { observer, useComputed } from "mobx-react-lite/dist/macro";
```

Todos:
* Integrate into the build process
Currently I'm using a custom tsconfig.json file (and the code itself is in JS not TS). 
* Make import accessible as `mobx-react-lite/macro`
I din't really know what the best solution woudl be here, maybe have a `macro.js` in the root directory that basically just forwards the export (`module.exports = require('./dist/macro')`)? But then we are still missing the Typescript definitions file.


build-rollup.js
```diff
@@ -86,13 +88,20 @@ function build(target, mode, filename) {
...
+}
+
 const main = async () => {
     await build("browser", "umd", "index.js")
     await build("browser", "umd.min", "index.min.js")
     await build("browser", "es", "index.module.js")
     await build("native", "cjs", "native.js")
     await build("custom", "umd", "custom.js")
     await build("custom", "es", "custom.module.js")
+    await build("custom", "es", "custom.module.js")
```
FredyC: I suppose this is duplicated by accident :)

package.json
```diff
@@ -31,15 +31,19 @@
   "homepage": "https://mobx-react.js.org",
   "peerDependencies": {
     "mobx": "^4.0.0 || ^5.0.0",
-    "react": "^16.8.0"
+    "react": "^16.8.0",
+    "babel-plugin-macros": "^2.6.1"
```
FredyC: I think it's better to document this instead of annoying everyone with peer dependency warning.

bernhardc:
OK, I think I found a good solution for the `mobx-react-lite/macro` import: a `/macro/index.js` and `/macro/index.d.ts` that just "forward" to the `/dist` directory.

FredyC:
Sounds good. Would you also write some paragraph to readme about it, please?

FredyC:
So I've replaced Rollup with TSDX, so this will need more tweaking. Probably add some postbuild script to tackle it.

FredyC:
Thanks for the docs update. I am thinking we should probably include some basic tests for this. Would you mind?

FredyC:
So, I am going to be releasing 2.x out soon, most likely by the end of the week. If you can manage to write the test, we can still squeeze it in the 1.x. Given that 2.x is not that big breaking change for most people, the 1.x won't be supported after release. However, it will require a tweaking build process toward TSDX for this PR as I've mentioned before.

bernhardc:
I don't think I have time this week. But no problem, I can migrate the build to TSDX.

FredyC:
@bernhardc Sorry for a delay here. I thought that 2.0 will come out soon, but with revelation of Concurrent mode, I decided to postpone release until it's completely figured out.

Either way, I've just merged TSDX changes to the master. Could you possibly rebase and update code?

FredyC:
If someone ever picks this up again, it will be better to start from scratch.

Issue #191, opened 2019-08-02
need support case of 'parent / childs render in the right order'


snadn:
is there any plan to support case of 'parent / childs render in the right order' https://github.com/mobxjs/mobx-react-lite/blob/master/test/observer.test.tsx#L600

FredyC:
Well, those tests were basically copied over from `mobx-react` and for some reason, it's commented out. Chances are it's actually working, I don't know. Can you present some real use case with reproduction, please?

snadn:
@FredyC my use case is same with the test. when logout, user data become null, and the components use user data which only render when login and control by parent get error.

FredyC:
Sorry, I am not sure I follow. That test is somewhat weirdly written and doesn't make much sense to me. I tried to rewrite it and it seems to be working. Please tweak that sandbox to reproduce your specific issue.

https://codesandbox.io/s/stupefied-pascal-o6wew

mweststrate:
@snadn for verification purposes, you could check if the problem disappears when using observer through mobx-react (mobx-react repackages mobx-react-lite + more goodies, so the rest of the impl will stay the same)

Please note that this issue is a duplicate of #153, so most probably not fixed yet .

As work around, doing the following in your app code before rendering React for the first time should also fix it: 

```javascript
import { configure } from "mobx"
import { unstable_batchedUpdates  } from "react-dom" // or react-native

configure({ reactionScheduler: unstable_batchedUpdates })
```

[source](https://github.com/mobxjs/mobx-react/blob/master/src/index.js)

snadn:
@mweststrate thanks, I will check later. I use mobx-react and mobx-react-lite in my project. I came across this problem when I use mobx-react-lite, and no problem when use mobx-react.

snadn:
@FredyC I reproduced this problem with mobx@4.13.0 mobx-react-lite@1.4.1 and async action.
https://codesandbox.io/s/brave-khayyam-we8ds
It's ok with mobx@4.13.0 mobx-react@6.1.1.
https://codesandbox.io/s/awesome-merkle-08slt

FredyC:
Ok, thanks for repro, that confirms it's indeed duplicate of #153 so closing here.

FredyC:
Although, I don't follow the use case. It seems that when I remove that `setTimeout` from the action, it works just fine in both cases (without batchedUpdates). Can you please elaborate on what is the use case for delaying action?

snadn:
[...]

PR #214, opened 2019-08-24
Opt-in variant of batched updates

### LINKED ISSUE #153: Add support for `unstable_batchedUpdates` when used with react-dom / react-native
**The issue:**
I suspect mobx-react-lite observer is handling actions/transactions and subsequent re-rendering differently to mobx-react. I don't know if the behaviour is intended and i am doing something wrong, or if its a bug so i would like to clarify this.

**Expected behaviour (the mobx-react way):**
- When an action is modifying the state of one or more models, all changes should be executed before an update (and re-render) is triggered on any affected components
```
export const ParentModel = types.model({
    title: "Parent",
    toggler: false,
    date: "",
    child: types.optional(ChildModel, {})
}).actions(self => ({
    toggleWithSideEffect(){
        self.child.date = new Date().toString();
        self.date = new Date().toString();
        self.toggler = !self.toggler; 
        // no rendering should have been done until this point  
        console.log("i toggle")
    },
}));
```

**Actual behavior (the mobx-react-lite way)**
- Certain combinations of state modifications cause updates to the state and re-render components in the middle of an action execution (transaction)
``` 
export const ParentModel = types.model({
    title: "Parent",
    toggler: false,
    date: "",
    child: types.optional(ChildModel, {})
}).actions(self => ({
    toggleWithSideEffect(){
        // child component referencing this value gets re-rendered after this statement
        self.child.date = new Date().toString();       
        self.date = new Date().toString();
        self.toggler = !self.toggler;
        // no rendering should have been done until this point
        console.log("i toggle")        
    },
}));
```

I created a (misspelled) [**repo**](https://github.com/gsabanov/mobscontext) that demonstrates the issue.

If you run the app in that repo, clicking on the toggle button will work the first time, but will crash the second time because the child model is using a computed value dependent on the **toggler** value in the parent model, that becomes null on the second iteration, and the child component re-renders before the parent. Running the same code with mobx-react observer behaves as expected, the child component doesn't even try to render when **toggler** is false.

The issue got weirder while i was composing the repo to test this behaviour.
The idea was to have two react apps next to each other, one with mobx-react one with mobx-react-lite and compare the difference, but loading mobx-react next to mobx-react-lite seems to correct the behaviour, even by just referencing a component that uses the mobx-react observer, see **[here](https://github.com/gsabanov/mobscontext/blob/90f406bf1afc156b569be9fe88d58d586cad37ba/src/index.js#L4)**
I threw in a useContext hook in my test as well to make sure the mobx-react observer isn't overwriting the mobx-react-lite observer when referenced.
Its as if loading mobx-react next to mobx-react-lite enchances/corrects transactional behavior of mobx-react-lite.

The example uses mobx-state-tree, but i was able to produce the same error with a couple of regular mobx nested models.

Here are two branches of the same repo with almost identical code, one with [mobx-react](https://github.com/gsabanov/mobscontext/tree/feature/just-mobx) and one with [mobx-react-lite](https://github.com/gsabanov/mobscontext/tree/feature/just-mobx-lite).

FredyC:
Kudos to @eps1lon for an idea https://github.com/facebook/react/issues/15080#issuecomment-524531411

Instead of creating two bundles per environment (dom, react-native), this solution is opt-in and as such should be causing less trouble for people who are not concerned by it. The bundles still exist although they are the same. The 2.0 will consolidate that.

Ultimately, the `mobx-react` could utilize the same approach. It could be just added in V6 and somehow deprecate native bundle.

Fixes #153


FredyC:
I tried to recover commented out test mentioned in https://github.com/mobxjs/mobx-react-lite/issues/191#issue-476035536 and it passes even without batched updates. I have probably misunderstood something about that use case.

FredyC:
@mweststrate Do you have any objections about this solution? Also please see my last comment, something doesn't add up.

mweststrate:
@FredyC, It seems there is still an react-dom dep in `src/optimizeScheduler.ts `?

But yes, I think the general approach is great! I think just exposing a `setReactBatchScheduler(fn)` would already do the trick, if it is documented that either `unstableBatchedUpdates` from react-dom or react-native should be passed in there before starting the React rendering.

It could be indeed that the test passes already, but that is mere lucky chance in that case about in which other the listeners are established, rather than that it is guaranteed to work. The general outline of the problem is:

1. two components, a parent and a child, listen to the same observable
2. the observable is changed *outside* a react event handler / `act` block
3. parent must re-render before child

Now the notification system might by mere change already notify in the right order (I think there is a Set nowaydays storing the listeners). However, it is not guaranteed. BatchedUpdates guarantees that even if a child is notified to render before a parent, the parent renders first nonetheless.

FredyC:
> It seems there is still an react-dom dep in `src/optimizeScheduler.ts `?

It's only for the type, it's stripped away in a bundle.

I am exporting `optimizeScheduler` from the bundle so it can be used directly if someone wants. Those separate import files are for convenience, but it's true I can document the existence of that function too.

> 2\. the observable is changed _outside_ a react event handler / `act` block

Yea, that's probably what I messed up, will try to fiddle with that.

xaviergonz:
what if it first try-catch require react-dom, if that failed try-catch require react-native and if that failed threw an exception?

xaviergonz:
Just checked, it does work, but it shows a warning on the console and a warning on compilation with create-react-app that such module cannot be found

mweststrate:
Yeah, I'd expect bundlers (and other analysis tools) to trip over it

FredyC:
Ok, I tried to break that test to stay as close to original as possible, but nada. It's kinda rare trying to break something and it doesn't want to be :)

Oh well, no point delaying it because of that. Let's leave it for another day when it becomes an actual problem.

PR #221, opened 2019-09-08
avoid some refs never being removed from the set

xaviergonz:


xaviergonz:
@RoystonS could you take a look?

FredyC:
Please rebase to "next", I've figured out conflicts, so there shouldn't be any `yarn.lock` changes in this PR.

RoystonS:
Hi. Sorry, only just spotted this. Can we add a unit test for this?  If there's a problem that's occurring in the wild, we should add a test to prevent regression.

FredyC:
@xaviergonz I agree we should add a test. I am even thinking if this isn't an actual fix for React Native Fast Refresh ... https://github.com/mobxjs/mobx-react-lite/issues/226#issuecomment-541727663. Fix for the "master" was found and it seems to be it's included in the "next" as well, but apparently there is something else wonky happening. Frankly, I am kinda lost in it.

FredyC:
Still waiting for tests here. If anyone has an idea how to test that scenario, help is definitely welcome. It's probably some edge case that won't be discovered without careful memory profiling.

FredyC:
I guess it's time to close. Without some tests, it doesn't make much sense to follow through here.

PR #222, opened 2019-09-08
Use ESLint instead of TSLint

FredyC:
I've decided not to bother with ESLint for 1.x and this will go straight to 2.0.

Also added running linter as part of CircleCI.


FredyC:
Huh, why would CircleCI fail so miserably? I haven't changed that much to break it so badly.

FredyC:
Replaced by #258 directly to master branch.

Issue #238, opened 2019-11-12
optimizeForReactDom/Native does not work in IE11 due to destructuring


Tarpsvo:
Hi!

As the `optimizeForReactDom`and `optimizeForReactNative` opt-ins are not transpiled and use object destructuring, they do not work in certain browsers (namely, IE11).

Can we avoid destructuring and just do it a bit more conventionally?

For example:
```
const ReactDOM = require('react-dom');
require('./dist').optimizeScheduler(ReactDOM.unstable_batchedUpdates);
```

FredyC:
Thanks for the report, you are right. Are you willing to provide PR?

Tarpsvo:
Definitely, provided a PR at #239.

FredyC:
Thank you. Can you also make the same PR to the `next` branch, please? It might be released soon as 2.0.

Tarpsvo:
Thanks! PR opened at #240.

FredyC:
I thank you :) Published 1.5.1.

Tarpsvo:
No problem! Thanks for the quick reply, merge and new release!

PR #239, opened 2019-11-12
Remove object destructuring from optimizeForReactDOM/Native

Tarpsvo:
Replaced object destructuring with a more conventional approach to fix issue #238.

It was crucial for `optimizeForReactDom.js` to provide support for IE11, but I also applied the same logic to `optimizeForReactNative.js` just to keep it consistent between the two files.


FredyC:
Thanks.

Issue #243, opened 2019-11-24
Keep generic arguments when using `observer`


Kukkimonsuta:
When using `observer` to wrap component that accepts generic argument the generic argument is lost effectively removing option to have observer components with generics. I've been able to hack together following workaround, but it would be great if generics weren't lost by default in `observer`. `mobx-react` works with generic component correctly.

Workaround:
```ts
function genericObserver<T extends React.FunctionComponent<P>, P extends object = T extends React.FunctionComponent<infer P> ? P : unknown>(fn: T): T {
	return observer<P>(fn) as T;
}
```

Current behavior:
![image](https://user-images.githubusercontent.com/737093/69500055-ffd3c100-0ef7-11ea-8b62-4d1f16616e75.png)

Desired behavior (similar to `mobx-react`):
![image](https://user-images.githubusercontent.com/737093/69500187-39f19280-0ef9-11ea-9a25-27840d5c6450.png)

Sample codesandbox: https://codesandbox.io/s/zen-mestorf-4hp7q

FredyC:
I am confused, looking at your CodeSandbox, how is the outcome different? In both cases, I see that `value` is `unknown`.

```tsx
const Bar = observer(<T extends unknown>({ value }: { value: T }) => {
  return <>{JSON.stringify(value)}</>;
});

const Baz = genericObserver(<T extends unknown>({ value }: { value: T }) => {
  return <>{JSON.stringify(value)}</>;
});
```

> `mobx-react` works with generic component correctly.

That is really strange thing to say because `mobx-react@6` builds on top of this lib and [doesn't do anything special about generics](https://github.com/mobxjs/mobx-react/blob/78c6b4964c7703de7a673913d759e9a2aab669f3/src/index.d.ts#L23). On the contrary it works even worse. Try to switch for `mobx-react` in your sandbox, the `<P>` won't work anymore.

Kukkimonsuta:
Within the component itself the typing is fine, however outside (after `observer` is applied) the generic parameter gets replaced by `unknown` and cannot be inferred.

![image](https://user-images.githubusercontent.com/737093/69500463-8c33b300-0efb-11ea-8b73-e62a898e64b1.png)

vs

![image](https://user-images.githubusercontent.com/737093/69500454-745c2f00-0efb-11ea-90b8-92f6a4d9a7ae.png)

These screenshots are taken directly from the sandbox.


I suspect `mobx-react` works because the typing for `observer` is different:

```ts
[...]

PR #244, opened 2019-11-25
Preserve generics when using `observer`

Kukkimonsuta:
Related issue https://github.com/mobxjs/mobx-react-lite/issues/243

This change allows wrapping generic components while preserving generic parameters.

There is also unintended change that preserves all static types. This has mostly positive effect, though `$$typeof`, `type`, `compare`, and `render` (which are not copied over) remain with their original type if defined on base component. I didn't find a way to omit these properties as any kind of narrowing will cause generic parameters to be converted to `unknown`.

### DIFF
--- a/src/observer.ts
+++ b/src/observer.ts
@@ -6,10 +6,16 @@ export interface IObserverOptions {
     readonly forwardRef?: boolean
 }
 
+export function observer<P extends object, TRef = {}>(
+    baseComponent: React.RefForwardingComponent<TRef, P>,
+    options: IObserverOptions & { forwardRef: true }
+): React.MemoExoticComponent<
+    React.ForwardRefExoticComponent<React.PropsWithoutRef<P> & React.RefAttributes<TRef>>
+>
 export function observer<
     C extends React.RefForwardingComponent<TRef, P>,
     P extends object = C extends React.RefForwardingComponent<any, infer T> ? T : unknown,
-    TRef extends object = C extends React.RefForwardingComponent<infer T, any> ? T : unknown
+    TRef = C extends React.RefForwardingComponent<infer T, any> ? T : unknown
 >(
     baseComponent: C,
     options: IObserverOptions & { forwardRef: true }
@@ -17,6 +23,11 @@ export function observer<
     React.MemoExoticComponent<
         React.ForwardRefExoticComponent<React.PropsWithoutRef<P> & React.RefAttributes<TRef>>
     >
+
+export function observer<P extends object>(
+    baseComponent: React.FunctionComponent<P>,
+    options?: IObserverOptions & { forwardRef: false }
+): React.FunctionComponent<P>
 export function observer<
     C extends React.FunctionComponent<P>,
     P extends object = C extends React.FunctionComponent<infer T> ? T : unknown
--- a/src/observer.ts
+++ b/src/observer.ts
@@ -6,16 +6,25 @@ export interface IObserverOptions {
     readonly forwardRef?: boolean
 }
 
-export function observer<P extends object, TRef = {}>(
-    baseComponent: React.RefForwardingComponent<TRef, P>,
+export function observer<
+    C extends React.RefForwardingComponent<TRef, P>,
+    P extends object = C extends React.RefForwardingComponent<any, infer T> ? T : unknown,
+    TRef extends object = C extends React.RefForwardingComponent<infer T, any> ? T : unknown
+>(
+    baseComponent: C,
     options: IObserverOptions & { forwardRef: true }
-): React.MemoExoticComponent<
-    React.ForwardRefExoticComponent<React.PropsWithoutRef<P> & React.RefAttributes<TRef>>
->
-export function observer<P extends object>(
-    baseComponent: React.FunctionComponent<P>,
-    options?: IObserverOptions
-): React.FunctionComponent<P>
+): C &
+    React.MemoExoticComponent<
+        React.ForwardRefExoticComponent<React.PropsWithoutRef<P> & React.RefAttributes<TRef>>
+    >
+export function observer<
+    C extends React.FunctionComponent<P>,
+    P extends object = C extends React.FunctionComponent<infer T> ? T : unknown
+>(
+    baseComponent: C,
+    options?: IObserverOptions & { forwardRef: false }
+): C & React.FunctionComponent<P>
+
 // n.b. base case is not used for actual typings or exported in the typing files
 export function observer<P extends object, TRef = {}>(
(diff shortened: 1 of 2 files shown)


FredyC:
@xaviergonz  @mweststrate Could really use your help with reviewing this, I am not that strong with TypeScript yet.

mweststrate:
I'll try to take a look tomorrow.

Kukkimonsuta:
I have noticed an issue where empty options were not accepted, which while silly could be an issue in future if more options are added. Should be fixed in last commit.

mweststrate:
The types are getting fairly complicated / overloaded now, I wonder if we can make it a bit simpler by creating one overload, and inferring return type roughly as follows, I think the type inference and error messages will be better if we do it this way, as it avoids 'backtracking' the generic arguments, and takes the input argument as starting point:

```typescript
export function observer<
  T extends React.FunctionComponent<any> | React.RefForwardingComponent<any>, 
  Options extends IObserverOptions
>(
    baseComponent: C,
    options?: Options,
): Options extends { forwardRef: true } 
  ? C extends React.RefForwardingComponent<infer TRef, infer P> 
    ? React.ForwardRefExoticComponent<React.PropsWithoutRef<P> & React.RefAttributes<TRef>>
    : never /* forwardRef set for a non forwarding component */
  : C
```

FredyC:
As Michel said...
> I wonder if we can make it a bit simpler by avoiding overloads

@Kukkimonsuta Did you try to remove other overloads?

Kukkimonsuta:
I have merged the two new overloads using Michels sample with slight modifications, however I'm not certain whether we want to remove the original overloads too as it would be breaking change and following transformation in users code would have to be done by users during migration to new version:

```ts
interface IProps {
    value: string;
}
const TestComponent = observer<IProps>(({ value, children }) => {
    return null
})
```

to

```ts
interface IProps {
    value: string;
}
const TestComponent = observer(({ value, children }: React.PropsWithChildren<IProps>) => {
    return null
})
```

mweststrate:
Hmmm, I've never seen the first pattern in the wild, personally. I think
the typical thing to do is to add `children: React.ReactNode` to the
`IProps`, especially because they don't have to be necessarily react nodes.

Personally I'd release it as minor, and rollback or add the additional
overload if it is an obstacle for many.

src/observer.ts
```diff
@@ -12,10 +12,29 @@ export function observer<P extends object, TRef = {}>(
...
     React.ForwardRefExoticComponent<React.PropsWithoutRef<P> & React.RefAttributes<TRef>>
 >
+
 export function observer<P extends object>(
     baseComponent: React.FunctionComponent<P>,
     options?: IObserverOptions
 ): React.FunctionComponent<P>
+
+export function observer<
+    C extends React.FunctionComponent<any> | React.RefForwardingComponent<any>,
```
FredyC: I am probably missing something here, but why can't we have `P extends object` here as well and then...

```suggestion
    C extends React.FunctionComponent<P> | React.RefForwardingComponent<unknown>,
```

Not sure about `unknown` if it can cause some issue, but sounds better than `any` imo.

FredyC:
Btw, this `next` branch, we can afford to have some breaking changes, there is very few of those so far. But I personally want to keep `observer<Props>` for sure.

src/observer.ts
mweststrate: the reason we have `any` here is that we want to prevent TypeScript to 'guess' a P (which often ends up in an `any` if it can't if there are generics etc), rather we compute it at the end (`infer P`) at the return position, based on the component type that was passed in. This approach is in my experience much more robust

src/observer.ts
```diff
@@ -12,10 +12,29 @@ export function observer<P extends object, TRef = {}>(
...
+): Options extends { forwardRef: true }
+    ? C extends React.RefForwardingComponent<infer TRef, infer P>
+        ? C &
+              React.MemoExoticComponent<
+                  React.ForwardRefExoticComponent<
+                      React.PropsWithoutRef<P> & React.RefAttributes<TRef>
+                  >
+              >
+        : never /* forwardRef set for a non forwarding component */
+    : C & React.FunctionComponent
```
mweststrate: Does `React.FunctionComponent` only declare the static members, or also the function taking props itself? As in the last case it would result in an intersection of `fn(props: P) & fn(props: any)` which is effectively `fn(props: any)`? If you need to pass in the generic `P`, you might need an additional conditional type and `infer P` on `C`

mweststrate: If I'm confusing or you're stuck, let me know and I'll take a look myself. If the type tests are completely to uncover the original issue, it should be fine.

FredyC:
I am usually doing this and we have various aliases for children (`NoChildren`, `SingleChild`, `Required<SomeChildren>`, etc...).

```ts
type TProps = SomeChildren & {
  extra: string
}

export const Comp = observer<TProps>(({ extra }) => { ... })
```

I suppose doing `observer(({ extra }: TProps) => { ... })` is not too horrible, but it will be pain to refactor that. We should probably create a [codemod](https://github.com/tusharmath/ts-codemod) or something to migrate this automatically.

The "problem" with official `PropsWithChildren` is it refers to `ReactNode` which includes `undefined` and that is wrong (an app will crash). It's some legacy crap going deep into how class components work. I have a custom ReactNode as using only FC and it works nicely.

mweststrate:
ok, maybe this pattern is more common than anticipated, and we should
continue support it, if 2/3 of the people in this thread already use it
:-P. It's pretty as well. I don't think we have it somewhere documented
explicitly, don't we?

FredyC:
It actually is documented ... https://mobx-react.js.org/observer-hoc
No surprise there since I wrote that docs 😎

mweststrate:
Hahaha cool!

src/observer.ts
Kukkimonsuta: It effectively declares only static members because the default is object `FunctionComponent<P = {}>`, so it doesn't get intersected with `any`, but passing actual props would work as well.

Kukkimonsuta:
I've experimented a bit further with this, however at this point I think it's out of my ability to merge rest of the overloads into one declaration.

The main issue I see is that there is a conflict of typing authority between the two: `observer<MyProps>((props) => null)` needs `observer` to declare type of the function component passed in, however `observer(<T extends unknown>(props: MyProps<T>) => null)` relies on `observer` respecting whatever is passed in.

mweststrate:
Sorry for the late follow up! Completely dropped the ball here, but I think it is fine as it is!

FredyC:
Ok, since this is going to `next` branch we can afford bit experimenting. I will publish another prerelease later with this.

PR #245, opened 2019-12-04
Remove unnecessary import from example in order to avoid confusion

grifotv:


FredyC:
Thanks for the fix, but note that `useObservable` is deprecated and will be removed in 2.0 very soon.

PR #246, opened 2019-12-14
useMemo instead of useState

hosseinmd:
UseMemo is the React standard to run a function and return a computed value in the function component, UseState in work too but it is not best practice.
This change was tested and worked well.
Great regards


joenoon:
This change would not be right.  useMemo can throw the value away if it chooses to and later re-run it, when what is needed here is actually a one-time-only value kept in state - exactly what it does.

FredyC:
Ehm, I did not notice this nice paragraph in docs? :)

image

So yea, if you are using `useMemo` for anything but optimization, chances are you are introducing bugs that can surprise you in the future.

PR #249, opened 2020-01-08
Update issue templates

FredyC:


FredyC:
Uh, wait, no, how is a PR? 😮 I was just fiddling how that templating works so I can set it up over at mobx repo first where I don't have access to this generator tool.

I will close it here for now and want to eventually figure out org-wide configuration.

PR #257, opened 2020-02-15
Use TSDX for bundling

FredyC:
Publishing 2.0 will probably take a while due to figuring out Concurrent mode and there is no community pressure to have it solved right now.

I have extracted TSDX changes to make it part of 1.x just to keep it sane across repos. Same as with mobx-react, there shouldn't be any breaking changes.


FredyC:
I suppose there is not much of surprise here, so let's just roll with it :)

PR #272, opened 2020-04-07
Fix losing props for observer component

vlazh:
Some code in my project has broken after latest updates because resulting observer component type loose original props. Take a look at [example](https://www.typescriptlang.org/play/?jsx=2#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wG4BYAKCqQA9JY5gA7GJKTdJOASQHkARgGd2AN3Z8wMYBGZC4AbypwVcIigAmsgDYBPOJmgB3FFA3JMAfgBccARAjbUzCpQC+VGvWjwNSNNqm3JgArswYMsxwEMJi7AA8AApwdGzMGvIxAFb+MAA0cAAqFnAAvIpuAHwAFAIoIgDCuJDMSKy2yOgwAHQWAGLGphosAOZN4LJtMPHFSJgFiZUFEFKRQrb8sVDiUJLSsvIAZIrKqoZQJmYWtjBQIUiubgCUHcQ9ALJIIBAAorQQ0jQ4xaU3inQw3QGFyGFj+AOAQOak1YYLe3USODAQgA6sAYAALCAhGAWJKVODHcE9CwAQRgt2AAmJSCEMwslQ5ri8DF8-kCRAMYQismiWx2SRStDSGVFOQwNTqjSRrXaiDRfSF+2YwOR00Wy1WBxsvEEIm2EkNcheaq6kM1kR1Kr1lS5lDoPLgfgCQUF4S1orN4oakul8ipdr9DuVoJQzF05IAPjaIf1BmZRo6Y3GlnA9msQ20ZZtAxatUIFfUkJnVQ0DWXjXmDtbG3IC+l5EpKKoDGnzHMbncHlQ3HBLHBg6lC2G0anoenmGNoyiWJh2EULAUV2vFqPxxTkx8vr9-oDq9Nw1DLn3MHDT0vz2iMSscXjCcTSTvKTO5nSGUy2KysyYBy5K2K0OynCotjBl+toapGshnq43YAPQoXA0FtjKF72oh97xFuUBwDuY4wQeEbCtq+E7mBSA7K6VAsGwHBcHANJPliJxdqoKBgSEIACOwjyeJQoQIVENLVAocAoHAbi2OxmJCNaABSADKAAa3Q-E4IBTFx3ZEDAIRQFE8TDKIlQKCgbjxChFkusOIloAc8CmnEUA0mUAYedUNJPK6LlyPARBCCE2gwAAjN58ReShjmUEFQghSy4UwAATDF7nmp5cDxaQQA).

Correct type checking:
image

Incorrect type checking (and children prop has appered):
image

Replace code:
image

And type checking is correct:
image

### DIFF
--- a/src/observer.ts
+++ b/src/observer.ts
@@ -34,7 +34,7 @@ export function observer<
                   >
               >
         : never /* forwardRef set for a non forwarding component */
-    : C & React.FunctionComponent
+    : C extends React.FunctionComponent<infer P> ? C & React.FunctionComponent<P> : never /* observer uses for non functional component */
 
 // n.b. base case is not used for actual typings or exported in the typing files
 export function observer<P extends object, TRef = {}>(
--- a/src/observer.ts
+++ b/src/observer.ts
@@ -7,6 +7,18 @@ export interface IObserverOptions {
     readonly forwardRef?: boolean
 }
 
+export function observer<P extends object, TRef = {}>(
+    baseComponent: React.RefForwardingComponent<TRef, P>,
+    options: IObserverOptions & { forwardRef: true }
+): React.MemoExoticComponent<
+    React.ForwardRefExoticComponent<React.PropsWithoutRef<P> & React.RefAttributes<TRef>>
+>
+
+export function observer<P extends object>(
+    baseComponent: React.FunctionComponent<P>,
+    options?: IObserverOptions
+): React.FunctionComponent<P>
+
 export function observer<
     C extends React.FunctionComponent<any> | React.RefForwardingComponent<any>,
     Options extends IObserverOptions
@@ -15,13 +27,14 @@ export function observer<
     options?: Options
 ): Options extends { forwardRef: true }
     ? C extends React.RefForwardingComponent<infer TRef, infer P>
-        ? React.MemoExoticComponent<
-              React.ForwardRefExoticComponent<
-                  React.PropsWithoutRef<P> & React.RefAttributes<TRef>
+        ? C &
+              React.MemoExoticComponent<
+                  React.ForwardRefExoticComponent<
+                      React.PropsWithoutRef<P> & React.RefAttributes<TRef>
+                  >
               >
-          >
         : never /* forwardRef set for a non forwarding component */
-    : React.MemoExoticComponent<C>;
+    : C;
 
 // n.b. base case is not used for actual typings or exported in the typing files
 export function observer<P extends object, TRef = {}>(
--- a/src/observer.ts
+++ b/src/observer.ts
@@ -7,18 +7,6 @@ export interface IObserverOptions {
     readonly forwardRef?: boolean
 }
 
-export function observer<P extends object, TRef = {}>(
-    baseComponent: React.RefForwardingComponent<TRef, P>,
-    options: IObserverOptions & { forwardRef: true }
-): React.MemoExoticComponent<
-    React.ForwardRefExoticComponent<React.PropsWithoutRef<P> & React.RefAttributes<TRef>>
->
-
-export function observer<P extends object>(
-    baseComponent: React.FunctionComponent<P>,
-    options?: IObserverOptions
-): React.FunctionComponent<P>
-
 export function observer<
     C extends React.FunctionComponent<any> | React.RefForwardingComponent<any>,
     Options extends IObserverOptions
@@ -27,14 +15,13 @@ export function observer<
     options?: Options
 ): Options extends { forwardRef: true }
     ? C extends React.RefForwardingComponent<infer TRef, infer P>
-        ? C &
-              React.MemoExoticComponent<
-                  React.ForwardRefExoticComponent<
-                      React.PropsWithoutRef<P> & React.RefAttributes<TRef>
-                  >
+        ? React.MemoExoticComponent<
+              React.ForwardRefExoticComponent<
+                  React.PropsWithoutRef<P> & React.RefAttributes<TRef>
               >
+          >
         : never /* forwardRef set for a non forwarding component */
-    : C extends React.FunctionComponent<infer P> ? C & React.FunctionComponent<P> : never /* observer uses for non functional component */
+    : React.MemoExoticComponent<C>;
 
 // n.b. base case is not used for actual typings or exported in the typing files
 export function observer<P extends object, TRef = {}>(
--- a/src/observer.ts
+++ b/src/observer.ts
@@ -34,7 +34,7 @@ export function observer<
                   >
               >
         : never /* forwardRef set for a non forwarding component */
-    : C;
+    : C extends React.FunctionComponent<infer P> ? C & React.FunctionComponent<P> : never;
 
 // n.b. base case is not used for actual typings or exported in the typing files
 export function observer<P extends object, TRef = {}>(
--- a/test/observer.test.tsx
+++ b/test/observer.test.tsx
@@ -762,6 +762,23 @@ it("should preserve generic parameters when forwardRef", () => {
     // this test has no `expect` calls as it verifies whether such component compiles or not
 })
 
+it("should keep original props types", () => {
+    interface TestComponentProps {
+        a: number;
+    }
+
+    function TestComponent({ a }: TestComponentProps): JSX.Element | null {
+        return null;
+    }
+
+    const ObserverTestComponent = observer(TestComponent);
+
+    const element = React.createElement(ObserverTestComponent, { a: 1 });
+    render(element)
+
+    // this test has no `expect` calls as it verifies whether such component compiles or not
+})
+
 // describe("206 - @observer should produce usefull errors if it throws", () => {
 //     const data = mobx.observable({ x: 1 })
 //     let renderCount = 0


src/observer.ts
```diff
@@ -34,7 +34,7 @@ export function observer<
                   >
               >
         : never /* forwardRef set for a non forwarding component */
-    : C & React.FunctionComponent
+    : C extends React.FunctionComponent<infer P> ? C & React.FunctionComponent<P> : never /* observer uses for non functional component */
```
mweststrate, line 37: I think this should just read `C`, which is already a React.FunctionComponent<P> per line 23, so need to infer and repeat the props. I'm pretty sure the definition on line 17 should be moved below this one, as the typings of this function are more specific. On line 30 I don't think the `C &` can be dropped as well.

vlazh: I don't quite understand why all these definitions if one definition can handle both variants.
```ts
export declare function observer<C extends React.FunctionComponent<any> | React.RefForwardingComponent<any, any>, Options extends IObserverOptions>(baseComponent: C, options?: Options): Options extends { forwardRef: true; }
    ? C extends React.RefForwardingComponent<infer TRef, infer P> ? 
        React.MemoExoticComponent<React.ForwardRefExoticComponent<React.PropsWithoutRef<P> & React.RefAttributes<TRef>>> : never
    : C;
```

FredyC:
Your latest changes are not good, have a look at failing tests.

Besides, I've seen patterns like `const comp = observer<TProps, TRef>((props, ref) => {...})` which wouldn't be possible with your simplified approach.

vlazh:
Got it. Now just fixes props.

FredyC:
FIY the test is still failing ;) I don't think you can make those typings much smaller really. Try to follow ideas from Michel on how to improve what you had.

vlazh:
Seems like my first edition was right.

> I think this should just read C, which is already a React.FunctionComponent
It is error with that modification.
And other suggestions are not helped.

vlazh:
That's it! Tests are passed.

FredyC:
Ok, and can you please add a test that covers your use case? To avoid regressions in the future.

vlazh:
Yes sure!

vlazh:
Done.

FredyC:
Thanks, published 2.0.4.

seivan:
Sorry but, unless I'm missing a bigger picture here, this causes a regression by adding implicit children which is not right. 
This means all your components are now expecting children, and you won't get a compile error if you're adding children to a component that isn't expecting it. 

[Playground](https://www.typescriptlang.org/play/?jsx=2&ssl=22&ssc=1&pln=24&pc=1#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wG4BYAKCqQA9JY5gA7GJKTdJOASQHkARgGd2AN3Z8wMYBGZC4AbypwVcIigAmsgDYBPOJmgB3FFA3JMAfgBccARAjbUzCpQC+VGvWjwNSNNqm3JgArswYMsxwEMJi7AA8AApwdGzMGvIxAFb+MAA0cAAqFnAAvIpuAHwAFAIoIgDCuJDMSKy2yOgwAHQWAGLGphosAOZN4LJtMPHFSJgFiZUFEFKRQrb8sVDiUJLSsvIAZIrKqoZQJmYWtjBQIUiubgCUHcQ9ALJIIBAAorQQ0jQ4xaU3inQw3QGFyGFj+AOAQOak1YYLe3USODAQgA6sAYAALCAhGAWJKVODHcE9CwAQRgt2AAmJSCEMwslQ5ri8DF8-kCRAMYQismiWx2SRStDSGVFOQwNTqjSRrXaiDRfSF+2YwOR00Wy1WBxsvEEIm2EkNcheaq6kM1kR1Kr1lS5lDoPLgfgCQUF4S1orN4oakul8ipdr9DuVoJQzF05IAPjaIf1BmZRo6Y3GlnA9msQ20ZZtAxatUIFfUkJnVQ0DWXjXmDtbG3IC+l5EpKKoDGnzHMbncHlQ3HBLHBg6lC2G0anoenmGNoyiWJh2EULAUV2vFqPxxTkx8vr9-oDq9Nw1DLn3MHDT0vz2iMSscXjCcTSTvKTO5nSGUy2KysyYBy5K2K0OynCotjBl+toapGshnq43YAPQodBbYyhe9qIfe8RblAcA7mOMEHhGwranhO5gUgOzIaoaFFPi3AiGgsgaHA2gsEgBSaDKoBgFxaB4nAaD4sA2gaEQURGOJYlMPIRg4AukFwIxPAwPg8goERCAGDgeD4vSWLWGhIyviEAjdGxIAod8Ai0FkQh2TEtAALTqBgblcWwKFgCE2jaChABMADswVwG5ahICMRBCEIkRjp41CUCwbAcFwcA0k+WInF2qgoGBIQgAI7D0SoaFiRJUltMa4ZUgAchAfiqWhyDfOIom4CAUxCAUMAQHAIAoAA1twIk2WAEncMlw6tShomxnAKDaEIg1aERmIvgSDTiZJ0nxNlW3kgNdjcAEqBEFA3RwK4bFyPANJlHA1QKMtcBuLYR3Pk8ZTkp23ZEDAIRQFE8TDKIlQKCgbjxChEMunNKVUPdQjwKacRQE95QxCWUDVDSTyumhADKb6SXAzAAl14DTZFcCyQi+IKV1UBEBgACEcBUGh4w9aw0TEpx3FwMFAAsL18aMol7TVUSnQS3BgFtv2ncNY1MPARjQCN3Qowc8BxQFMA0gAjM9h3LaUCim1Uqng8AohqeS9soTSlT6w90VCMbNIReU8QY+aWNWzbdv5SoDtOyhHsR3AcNBzs7uqa6QA) 

The previous implementation would give you a compile error if you inserted children when they weren't expected which is more correct and this just removes more checks.

FredyC:
@seivan These types are already complicated as is. If you have some idea on how to cover a problem of this issue and yours, PR would be most welcome. This is too advanced for me.

seivan:
@mweststrate I think you're on to something with this [comment](https://github.com/mobxjs/mobx-react-lite/pull/272#discussion_r404694510) just using `C` will suffice. 

```jsx
const A = ({ a }: AProps) => {
    return <div>{a}</div>;
}
//Inferred Type Signature
//const ObserverA: ({ a }: AProps) => JSX.Element
const ObserverA = observer(A);


const B = ({ a, children }: React.PropsWithChildren<AProps>) => {
    return <div><h1>{a}</h1>{children}</div>;
}
//Inferred Type Signature
//const ObserverB: ({ a, children }: React.PropsWithChildren<AProps>) => JSX.Element
const ObserverB = observer(B);
```

seivan:
@FredyC I am suggesting to minimize them by removing the complexity. 

Replacing 
```typescript
C extends React.FunctionComponent<infer P> ? C & React.FunctionComponent<P> : never;
```

with

```typescript
    : C
```

as suggested by @mweststrate in [comment](https://github.com/mobxjs/mobx-react-lite/pull/272#discussion_r404694510)

But the reason I am writing here, is because I am not 100% sure if I understood the PR correctly and what it's solving. 
The impression I have is that @vlazh got his/her compile errors when passing children in on components that didn't have an explicit `children?: React.ReactNode` in their props.

FredyC:
The test @vlazh added should be fairly explanatory. I tried it without this change and this line would fail because `a` property is not present on the implicit Attributes (or something).

```ts
React.createElement(ObserverTestComponent, { a: 1 });
```

The best course of action would be if you can create a PR with your suggested change and add a test that covers your issue. Unless all tests are passing, we don't have a solution.

seivan:
@FredyC That's odd, my tests passed. 

What typescript version are you running?

seivan:
@FredyC Made a [PR](https://github.com/mobxjs/mobx-react-lite/pull/275) without changing the tests. 
The tests ran successful locally with with TS `3.8.3`.

vlazh:
I have no problem just with
```ts
: C
```
while it keep all original props.

Consider to use `React.NamedExoticComponent` instead of `{displayName: string}`.
```ts
C extends React.FunctionComponent<infer P> ? C & React.NamedExoticComponent<P> : never;
```

It might worth to change tests.

seivan:
Here's the thing, Typescript lets you declare properties instead of interfaces to conform to as well.

Instead of saying `C extends NamedExoticComponent ` and having to update that when that **also** gets deprecated and renamed for whatever reason. 

I can instead say `C has {displayName?: string}` and have it work none the less, instead of introducing more meaningless types, until React decides to change that name, but at the very least, it's one less thing to cover.

vlazh:
Using of `{displayName: string}` looks like dirty hack specially for tests. I prefer to use existing types if possible instead of introduce yet another type.  It's more suitable type which returnd from `React.memo` which uses in `observer`. Tomorrow, React might decides to extend the `NamedExoticComponent` and in that case we need nothing to do.

seivan:
Now, I could be wrong but as far as React is concerned, only `displayName` exists,  in fact, `NamedExoticComponent` is some [foreign](https://github.com/facebook/react/search?q=NamedExoticComponent&unscoped_q=NamedExoticComponent&type=Issues) interface stored on `DefinitelyTyped`. 


It's worth taking a step back before calling stuff hacks, because the reason why you had issues and these messes happen is because code like: 
```typescript
 ? C extends React.RefForwardingComponent<infer TRef, infer P>
        ? C &
              React.MemoExoticComponent<
                  React.ForwardRefExoticComponent<
                      React.PropsWithoutRef<P> & React.RefAttributes<TRef>
                  >
              >
        : never /* forwardRef set for a non forwarding component */
    : C extends React.FunctionComponent<infer P> ? C & React.FunctionComponent<P> : never;
```

There's nothing inherently wrong with it, it's just... that it exists. 


Now, we need **JUST** `displayName`, do we introduce `NamedExoticComponent` or do we outright say, that that `displayName` is part of the type? 

Think about this
We don't care that X conforms to `NamedExoticComponent`
We care that X has `{displayName: string}`, because that's what we're calling when we need the `displayName`

A hack is introducing 30 different granular interfaces because of some pipe dream about good abstractions.

vlazh:
> We care that X has {displayName: string}, because that's what we're calling when we need the displayName

Yes, and we already have specially type from react types lib for that case but you prefer write your own. May be it's absolutly no matter for you but   
> I prefer to use existing types if possible instead of introduce yet another type.

And relying on source code of `observer`
> It's more suitable type which returnd from React.memo which uses in observer. 

I don’t understand what you are arguing about. :)
In you PR you propose your solution for which I propose you to use a little bit another approach (better for my opinion) and you decide which you want use. Eventually last word for maintainers.

PR #278, opened 2020-04-18
Fix for a possible memory leak in observer

hzzcc:
Reaction cause memory leak

### Code change checklist

-   [ ] Added/updated unit tests
-   [ ] Updated `README` if applicable

### DIFF
--- a/src/useObserver.ts
+++ b/src/useObserver.ts
@@ -111,6 +111,7 @@ export function useObserver<T>(
 
         return () => {
             reactionTrackingRef.current!.reaction.dispose()
+            // @ts-ignore
             reactionTrackingRef.current!.reaction = null
             reactionTrackingRef.current = null
         }
--- a/src/useObserver.ts
+++ b/src/useObserver.ts
@@ -67,8 +67,9 @@ export function useObserver<T>(
                 // We haven't yet reached useEffect(), so we'll need to trigger a re-render
                 // when (and if) useEffect() arrives.  The easiest way to do that is just to
                 // drop our current reaction and allow useEffect() to recreate it.
-                reactionTrackingRef.current.reaction.dispose()
-                reactionTrackingRef.current.reaction = null
+                reactionTrackingRef.current!.reaction.dispose()
+                // @ts-ignore
+                reactionTrackingRef.current!.reaction = null
                 reactionTrackingRef.current = null
             }
         })
--- a/src/useObserver.ts
+++ b/src/useObserver.ts
@@ -67,7 +67,8 @@ export function useObserver<T>(
                 // We haven't yet reached useEffect(), so we'll need to trigger a re-render
                 // when (and if) useEffect() arrives.  The easiest way to do that is just to
                 // drop our current reaction and allow useEffect() to recreate it.
-                newReaction.dispose()
+                reactionTrackingRef.current.reaction.dispose()
+                reactionTrackingRef.current.reaction = null
                 reactionTrackingRef.current = null
             }
         })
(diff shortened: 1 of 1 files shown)


FredyC:
Please fix error in failing build. And also can you elaborate on why this change? Did you encounter some issue that you are able to reproduce? In that case, can you add a test that covers that issue?

mweststrate:
Thanks for taking the effort of creating this PR, but please elaborate on how this prevents a memory leak, as we can't accept changes without some rationally or verification

hzzcc:
image
image

FredyC:
Sorry, but one screenshot and `tsignore` (😮) in the code won't do. Can you explain the problem in words?

src/useObserver.ts
```diff
@@ -67,7 +67,9 @@ export function useObserver<T>(
                 // We haven't yet reached useEffect(), so we'll need to trigger a re-render
                 // when (and if) useEffect() arrives.  The easiest way to do that is just to
                 // drop our current reaction and allow useEffect() to recreate it.
-                newReaction.dispose()
+                reactionTrackingRef.current!.reaction.dispose()
+                // @ts-ignore
+                reactionTrackingRef.current!.reaction = null
```
FredyC, line 72: This just doesn't feel like it's necessary. I mean it won't probably harm anything, but I would rather keep code that makes sense only. It's obvious, that calling `dispose` won't remove reaction from the memory immediately. Even removing its reference doesn't give such assurance, it's up to GC.

src/useObserver.ts
```diff
@@ -67,7 +67,9 @@ export function useObserver<T>(
                 // We haven't yet reached useEffect(), so we'll need to trigger a re-render
                 // when (and if) useEffect() arrives.  The easiest way to do that is just to
                 // drop our current reaction and allow useEffect() to recreate it.
-                newReaction.dispose()
```
FredyC, line 70: Explain why this is harmful to have here, please.

hzzcc: I found that when keep "newReaction" in the func，the reference newReaction will still in the GC.
But when delete the newReaction, use reactionTrackingRef.current, the reference will be removed.

hzzcc: [https://github.com/hzzcc/mobx-react-lite-test](https://github.com/hzzcc/mobx-react-lite-test)
Here is the test repo.

hzzcc:
[https://github.com/hzzcc/mobx-react-lite-test](https://github.com/hzzcc/mobx-react-lite-test)
Here is the test repo.

hzzcc:
The test flow: 
1. click "onChange"
2. change the input
1. click "onChange" to hide input
3. profile the memory

FredyC:
Sorry, but the test repo is not gonna be enough. If you have discovered a flaw, we need to figure out to reproduce it in a test file and that needs to be added as a part of this PR. The reason is to avoid regressions for future changes.

src/useObserver.ts
FredyC: @hzzcc Can you respond to this? Not sure why you marked it as resolved without an answer.

hzzcc: @FredyC Yes, this is up to GC, but  can take heap snapshot by youself, click several times until memory stabilizes.

FredyC: I don't follow. How does that explain why do you need to set a reaction to null here?

hzzcc: Dereference 'reaction'. But, I guess input change event cause these. when I change to 'div', the problem disappear.

FredyC: Ok, so please remove this from PR if it's not necessary, it only brings confusion.

FredyC:
I will keep it open as it might worth investigating if there is an actual memory leak. Help is certainly wanted here.

FredyC:
We now have an alternative solution #332 so it's probably not worth it to worry about a memory leak in here. Besides nobody else seem to have noticed anything.

PR #279, opened 2020-04-21
fix(types): deprecated RefForwardingComponent usage

### LINKED ISSUE #263: Namespace 'React' has no exported member 'RefForwardingComponent'.
## Intended outcome

I'm building an application, and after doing some work with MobX, I ran a production build to ensure everything was working as expected, but it errored out. Development build does work exactly as intended, however.

## Actual outcome

I get the following error:

```
✖ ERROR ./node_modules/mobx-react-lite/dist/observer.d.ts
ERROR in ./node_modules/mobx-react-lite/dist/observer.d.ts(5,84):
TS2694: Namespace 'React' has no exported member 'RefForwardingComponent'.
```

## How to reproduce the issue

Creating a simple component will trigger the issue:

```
export const Example: FunctionalComponent<Issue> = observer((props: Issue) => {
    ...
});
```

Nothing more advanced than this needed. 

## Versions
```
"mobx": "^5.15.4",
"mobx-react-lite": "^1.5.2",
"@types/react": "16.9.19",
```

Browsing through @types/react, I did notice they deprecated `RefForwardingComponent`, but even going to an older version (`16.9.19`) didn't fix this problem. Strangely, if I edit `observer.d.ts` in `node_modules`, and import React there (`import React from 'react';`), the production build works just fine. Just something I noticed.

nulladdict:
Update react and react-dom types
Use ForwardRefRenderFunction instead of RefForwardingComponent

Closes #263


nulladdict:
I also ran prettier and fixed a small typo

There were no need to add/change tests, since the test present already check observer usage with and without `forwardRef` option

I did however had to change the base case typings, since TS would complain that `baseComponent: React.FunctionComponent<P>` overload is not compatible with the implementation

FredyC:
I don't think we should do this yet as it would block people with older versions of types package. As long as `RefForwardingComponent` exists, we should stick to it. 

Note that you only changed versions of DEV dependencies. We cannot enforce a specific version to consumers because of global types.

FredyC:
Hm, I see now I wrote a comment that PR fixing deprecation is welcome. My bad, that might have been premature. I am not sure how old that change is.

Thanks for a contribution anyway. Let's leave this open for now and when more complaints about the wrong Ref typing pile up, we can push this through.

nulladdict:
I think the changes dates back to Feb 20 according to [this PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/42398), so it's still pretty fresh and I don't mind leaving it on hold for now

Also, I don't think TS has anything to "fix" non-existing types other than [`typesVersions`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-1.html#version-selection-with-typesversions) property, which unfortunately relies on TS version

FredyC:
This PR has conflicts and I will soon move the codebase to mobx repo (https://github.com/mobxjs/mobx/issues/2530). I am still not sure we are ready to do this since the deprecated type is still available even in the latest types package version.

PR #280, opened 2020-04-21
wip: move render side effect to useEffect hook

nulladdict:
### Code change checklist

-   [ ] Added/updated unit tests

### DIFF
--- a/src/useAsObservableSource.ts
+++ b/src/useAsObservableSource.ts
@@ -24,8 +24,10 @@ export function useAsObservableSourceInternal<TSource>(
     ) {
         throw new Error(`the shape of objects passed to ${culprit} should be stable`)
     }
-    runInAction(() => {
-        Object.assign(res, current)
+    React.useEffect(() => {
+        runInAction(() => {
+            Object.assign(res, current)
+        })
     })
     return res
 }


nulladdict:
It seems to me that `useAsObservableSourceInternal` has a bug with hooks and early return.
If in the runtime `current` argument changes from plain object to `undefined` function will short-circuit and not run `useState` hook resulting in the rules of hooks violation

nulladdict:
If I move `useEffect` call above shape check to avoid rules of hooks problems I get 3 failed tests 2 of which are render count increases which are inevitable in this approach but probably fine, since side effect in render might lead to a bigger problems.

The last test is for `combining observer with props and stores › allows non-observables to be used if specified as as source` and there's an actual behaviour difference which I don't know if right or not

FredyC:
Wait, you cannot have useEffect after series of condition, **that's against rules of hooks** ;)

Either way, if you think there is some issue, please supply a new test to confirm it. Right now you have kinda broken it more than fixed something :)

nulladdict:
I think this is safe to close for know, because we only reproduction for this warning is a weird synchronised change to observable and not observable state

Issue #274, opened 2020-04-08
The `useAsObservableSource` is causing React warning


SidKwok:
# ✋✋✋ Issue was revived based on https://github.com/mobxjs/mobx-react-lite/issues/274#issuecomment-652958242

The following comment is left here for clarity but doesn't reflect an actual issue.

---

## Intended outcome

Not trigger warning.

## Actual outcome

warning from react:
```log
index.js:1 Warning: Cannot update a component (`wrappedComponent`) while rendering a different component (`wrappedComponent`). To locate the bad setState() call inside `wrappedComponent`, follow the stack trace as described in https://fb.me/setstate-in-render
    in wrappedComponent (created by Context.Consumer)
    in Route (at App.tsx:48)
    in Switch (at App.tsx:47)
    in div (at App.tsx:46)
    in App (at App.tsx:60)
    in Router (created by HashRouter)
    in HashRouter (at App.tsx:59)
    in Unknown (at src/index.tsx:7)
```

## How to reproduce the issue
Try to run this code
```jsx
import React from 'react';
import { observer, useLocalStore } from 'mobx-react-lite';
import { useHistory, Route, Switch, HashRouter } from 'react-router-dom';

const context = React.createContext(null);

const Com = observer(() => {
  const store = React.useContext(context) as any;
  return (
    <div>
      <p>{store.str}</p>
      <button onClick={store.go}>
        test
      </button>
    </div>
  );
});

const Page = observer(() => {
  const history = useHistory();
  const storeSource = {
    history,
    search: history.location.search
  };
  const store = useLocalStore(source => ({
    get str() {
      return source.search
    },
    go() {
      source.history.replace(
        '/test?a=1'
      );
    }
  }), storeSource);
  return (
    <context.Provider value={store as any}>
      <div>
        <Com />
      </div>
    </context.Provider>
  );
});

function App() {
  return (
    <div className="App">
      <Switch>
        <Route
          path="/test"
          component={Page}
          exact
        />
      </Switch>
    </div>
  );
}

export default () => (
  <HashRouter>
    <App />
  </HashRouter>
);

``

## Versions

react: 16.13.1
react-dom: 16.13.1
react-router-dom: 5.1.2
mobx: 5.15.4
mobx-react-lite: 2.0.4

FredyC:
Can you please use CodeSandbox to showcase the issue? It would easier for us to investigate more closely.

SidKwok:
[...]

PR #281, opened 2020-04-21
useLocalStore source change from something to nothing

nulladdict:
### Code change checklist

-   [x] Added/updated unit tests

Failing test for [this comment](https://github.com/mobxjs/mobx-react-lite/issues/274#issuecomment-617326964)

### DIFF
--- a/src/useAsObservableSource.ts
+++ b/src/useAsObservableSource.ts
@@ -8,20 +8,26 @@ export function useAsObservableSourceInternal<TSource>(
     usedByLocalStore: boolean
 ): TSource {
     const culprit = usedByLocalStore ? "useLocalStore" : "useAsObservableSource"
+    if (__DEV__ && usedByLocalStore) {
+        const [initialSource] = React.useState(current)
+        if (
+            (initialSource !== undefined && current === undefined) ||
+            (initialSource === undefined && current !== undefined)
+        ) {
+            throw new Error(`make sure you never pass \`undefined\` to ${culprit}`)
+        }
+    }
     if (usedByLocalStore && current === undefined) {
         return undefined as any
     }
-    if (process.env.NODE_ENV !== "production" && !isPlainObject(current)) {
+    if (__DEV__ && !isPlainObject(current)) {
         throw new Error(
             `${culprit} expects a plain object as ${usedByLocalStore ? "second" : "first"} argument`
         )
     }
 
     const [res] = React.useState(() => observable(current, {}, { deep: false }))
-    if (
-        process.env.NODE_ENV !== "production" &&
-        Object.keys(res).length !== Object.keys(current).length
-    ) {
+    if (__DEV__ && Object.keys(res).length !== Object.keys(current).length) {
         throw new Error(`the shape of objects passed to ${culprit} should be stable`)
     }
     runInAction(() => {
--- a/src/useLocalStore.ts
+++ b/src/useLocalStore.ts
@@ -4,14 +4,19 @@ import React from "react"
 import { useAsObservableSourceInternal } from "./useAsObservableSource"
 import { isPlainObject } from "./utils"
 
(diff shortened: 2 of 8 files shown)


FredyC:
@nulladdict Thanks and can you also add a fix for it? :)

nulladdict:
I've changed typings for `useLocalStore` so TS wouldn't allow mixing having and not having `source` argument

Fix for throwing correct error in this case looks kinda meh to me, but I couldn't figure out how to make it look better

Also I've encountered an interesting behaviour with `renderHook`, it throws on `Invariant Violation` even though the hook threw different error. So I wonder if better way is to let `useAsObservableSourceInternal` throw `Invariant Violation` and replace new error with warning

src/useAsObservableSource.ts
```diff
@@ -8,6 +8,15 @@ export function useAsObservableSourceInternal<TSource>(
     usedByLocalStore: boolean
 ): TSource {
     const culprit = usedByLocalStore ? "useLocalStore" : "useAsObservableSource"
+    if (process.env.NODE_ENV !== "production" && usedByLocalStore) {
```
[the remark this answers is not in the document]
nulladdict: I believe this is there so that tests can assert dev errors. (there's `global.__DEV__ = false` in a `jest.setup.js` file)

I'll change to `__DEV__` and make sure dev tests set that correctly

nulladdict:
I've pushed fixes, will squash later once everything is okay to proceed

src/useLocalStore.ts
```diff
@@ -4,14 +4,19 @@ import React from "react"
 import { useAsObservableSourceInternal } from "./useAsObservableSource"
 import { isPlainObject } from "./utils"
 
+export function useLocalStore<TStore extends Record<string, any>>(initializer: () => TStore): TStore
 export function useLocalStore<TStore extends Record<string, any>, TSource extends object = any>(
```
FredyC, line 8: Wouldn't it be better to require `TSource` of object type here? I mean to remove default for `any` (which can be undefined).

src/useLocalStore.ts
```diff
@@ -4,14 +4,19 @@ import React from "react"
...
     initializer: (source: TSource) => TStore,
+    current: TSource
+): TStore
+export function useLocalStore<TStore extends Record<string, any>, TSource extends object = any>(
```
FredyC, line 12: I am not sure I follow the reason for this last overload. Doesn't make it accept `undefined`?

nulladdict: According to the [docs](https://www.typescriptlang.org/docs/handbook/functions.html#overloads) last "overload" is not a part of overload list, it's what TS calls "implementation" or base case and it must implement all actual overloads

It's not used for actual typing and not exported in typing files, just like base case for `observer()`

While type-checking TS would use first [of two] "compiling" overload and not the base case, so:
- something like `useLocalStore(() => ({}), undefined)` would fail at second overload with `TS2345: Argument of type 'undefined' is not assignable to parameter of type 'object'.`
- something like `useLocalStore(() => ({}))` or `useLocalStore(({}) => ({}), {})` would compile picking first or second overload
- something like `useLocalStore(props => ({}))` would fail with `TS2345: Argument of type '(props: any) => {}' is not assignable to parameter of type '() => {}'.`
- and finally something like 
```ts
const source = { foo: "bar" } as { foo: string } | undefined
useLocalStore(
    props => ({
        props
    }),
    source
)
```
where source might be some object or `undefined` would fail with `TS2345: Argument of type '{ foo: string; } | undefined' is not assignable to parameter of type 'object'.   Type 'undefined' is not assignable to type 'object'.`

FredyC: Cool, I did not know about that. Thanks for the exhaustive explanation.

Issue #286, opened 2020-05-15
'Symbol' is undefined in IE11


MaartenStaa:
It seems the way [the `for of` loop in `cleanUncommittedReactions`](https://github.com/mobxjs/mobx-react-lite/blob/7045287a0d1588321c1789e18b410c4567283159/src/reactionCleanupTracking.ts#L86) is transpiled for publishing raises issues in IE11:

```js
for (var _iterator = uncommittedReactionRefs, _isArray = Array.isArray(_iterator), _i = 0, _iterator = _isArray ? _iterator : _iterator[Symbol.iterator]();;) {
```

## Intended outcome

Since I'm using MobX 4, I expected to be able to use the related MobX React and MobX React Lite libraries without issues in IE11.

## Actual outcome

An exception is thrown and logged in the console: `'Symbol' is undefined`. I imagine this also prevents the cleanup from actually happening.

## How to reproduce the issue

It seems CodeSandbox doesn't do transpilation for the code, so I couldn't test that one in IE11. But I believe this will happen in any app using `mobx-react-lite`.

## Versions

mobx: 4.15.4
mobx-react: 6.2.2
mobx-react-lite: 2.0.6

FredyC:
Thanks for the report, you are right. The `for..of` is about iterators which need a Symbol. Do you want to submit a PR to fix? It should be as simple as replacing with regular for loop over the `uncommittedReactionRefs.values()`.

PR #287, opened 2020-05-15
Get rid of for of loop

piotrferens:
fixes: #286


FredyC:
Cool, thanks. Forgot that `Set` has `forEach` so that was really easy fix.

FredyC:
Published 2.0.7

PR #299, opened 2020-08-06
Allow store modifications during rendering of components

### LINKED ISSUE #274: The `useAsObservableSource` is causing React warning

olee:
This change shifts forceUpdate calls triggered by useObservable reactions during rendering of a component into a useLayoutEffect to prevent calling setState on another component than the currently rendering one.

Fixes #274


FredyC:
@urugator I might be misunderstanding something, but wouldn't be a problematic with Concurrent mode to have global variables that assumes that only one component is currently rendered? I thought that's what CM is trying to solve to render multiple things at once.

urugator:
@FredyC

To run 2 functions truly concurrently you would need WebWorkers or something.
The question is whether react can start rendering other components from within your currently rendering component eg react would expose a hook like
```javascript
function useWeirdHook() {
  // run other components from within this one...
  reactGlobals.scheduledRenders.forEach(render => render())
}
```
I don't think this can happen, but I can't tell for sure.

@olee

Do we have any guarantee that the source is mutated (therefore the effects are invoked) inside `observer`? If not we might miss `isRendering` flag:

```javascript
function component() {
  // the mutation happens here
  // therefore the reaction is scheduled here
  // but we are not in observer yet
  // therefore it doesn't know we are rendering
  // and it will schedule update immediately
  const source = useAsObservableSource(); 

  return useObserver(() => { /* the flag is switched here */ })
}
```

olee:
@urugator 
It is not possible for react to render a component completely in parallel to another component.
The concurrency of concurrent mode is only in regard to a whole render loop.
Each fiber is still rendered synchronously and it can just pause rendering between fibers.

As for the second example you provided:
it would fail again as the capturing of forceUpdate calls happens only inside the useObserver callback.
However, I don't know if there is any good solution around that.

src/useQueuedForceUpdate.ts
```diff
@@ -0,0 +1,41 @@
+import React from "react"
+
+export type ForceUpdate = () => void
+
+let insideRender: boolean = false
+let forceUpdateQueue: ForceUpdate[] = []
+
+export function useQueuedForceUpdate(forceUpdate: ForceUpdate): ForceUpdate {
+    return () => {
+        if (insideRender) {
```
urugator, line 10: If the component invalidates itself, we don't have to schedule anything since the component si already being rendered with the late-most values. Can you add a check for that?

I was hoping we could do something similar for all components, but now I am not sure. The idea is that when component starts to rendering (before the `useLayoutEffect` has a chance to run) it would remove itself from `forceUpdateQueue`.

urugator: Or rather instead of *removing itself from queue* (as it would require some lookup), there would be a check in this else branch like 
```else if (!wasAlreadyRendered) {
  forceUpdate()
}
````

**Comment by FredyC** [comment]
> Each fiber is still rendered synchronously and it can just pause rendering between fibers.

I wonder, does that actually mean that when one component is paused, the other is rendering? That wouldn't cope well with any global state unless it's tied to fiber which we don't have access to.

I think it would be worth trying to run current tests with Concurrent mode (+experimental React versions) to see a true behavior. Switching to CM should be possible with [RTL custom render](https://testing-library.com/docs/react-testing-library/setup#custom-render).

File: src/useQueuedForceUpdate.ts [hunk 2]
```diff
@@ -0,0 +1,41 @@
...
+    // start intercepting force-update calls
+    insideRender = true
+    forceUpdateQueue = []
+    try {
+        const result = callback()
+
+        // stop intercepting force-update
+        insideRender = false
+        // store queue or nothing if it was empty to execute useLayoutEffect only when necessary
+        const queue = forceUpdateQueue.length > 0 ? forceUpdateQueue : undefined
```
urugator, line 28: Don't we need a copy here?

urugator:
@FredyC 
> that when one component is paused, the other is rendering

The way I understand it: You cannot pause the component in the middle of rendering. There is no way for react to tap into the component, unless you could explicitely call something like `React.yield()`.
But if you could call something like this anytime, react would probably not be able to make assumptions about the component state that are required for concurrency.
It doesn't matter which fiber is rendering, you simply cannot communicate between components (between fibers), therefore the warning about `setState`.
So if *anything* is rendering (no matter which fiber), we simply don't call `setState` (because we can't) and instead we defer these updates to `useLayoutEffect` (of the same component that is currently rendering and causing these updates).
If the fiber isn't commited, the `useLayoutEffect` won't get called so no problem here.
We use global var for queue and rendering flag, but they got re-setted for every render, and since we know that only one render can happen at a time, there is no concurrency.

olee:
@urugator this was really a perfect explanation on what this does - I couldn't have done this any better 😄 
Maybe we should even copy / link that explanation into the code where this logic is located at

src/useQueuedForceUpdate.ts
olee: I think this might be possible, but I'm not sure it wouldn't have any unwanted side-effects so I'd rather leave it as-is.
As this should also not happen that often and only affects a single component, I don't think this would matter much performance wise.

src/useQueuedForceUpdate.ts
olee: The copy is stored in the `queue` const and `forceUpdateQueue` is always initialized to an empty array before the next render so it should be fine.

urugator:
Could you add some test(s)? At least something simple that would demonstrate why it's needed.
Otherwise I am fine with it.
CircleCI seems to be somehow stuck??

Pinging @mweststrate in case he has something to add.

mweststrate:
@urugator yeah I want to study this in a bit more detail. Some tests are definitely needed. Also would be great if someone would test that this doesn't break with hot component reloading as set up in CRA

olee:
@urugator the thing is, I tried to write up a test, but failed for various reasons:
1. I never wrote such react tests
2. I have no idea how to test for those incorrect calls to setState during rendering of a component

Would you be able to write a test for that?

@mweststrate why would you think this could break with hot reloading?
At least I can confirm that in the project where I am using this fixed version hot reloading works without any issues if that answers your question.

mweststrate:
Ok, merging and releasing! I thinks should be good to go, and otherwise there is nothing we can't revert :)

mweststrate:
Released as 2.1.0. Thanks!

Issue #296, opened 2020-07-23
observerBatching presets always bundle CommonJS


davezuko:
Hey! Love this library, and wanted to let you know about something I discovered today while trimming down our bundles:

When using the side-effect imports to configure observer batching (e.g. `import "mobx-react-lite/batchingForReactDom"`), mobx-react-lite imports the CommonJS module regardless of the preferred module type. This is problematic because mobx-react-lite gets bundled twice unless you happen to be using CommonJS. The offending code:

```js
// batchingForReactDOM.js
const ReactDOM = require("react-dom")
require("./dist").observerBatching(ReactDOM.unstable_batchedUpdates)

// ./dist resolves to ./dist/index.js, which contains:
if (process.env.NODE_ENV === 'production') {
  module.exports = require('./mobxreactlite.cjs.production.min.js')
} else {
  module.exports = require('./mobxreactlite.cjs.development.js')
}
```

You can circumvent this issue by just configuring it yourself, which to me seems preferable anyways since it doesn't rely on side-effects or further complication to the packaging infrastructure.

```ts
import * as ReactDOM from "react-dom"
import {observerBatching} from "mobx-react-lite"
observerBatching(ReactDOM.unstable_batchedUpdates)
```

In my case, this change saved 3kb (~50%) from mobx-react-lite-related imports.
 
## Intended outcome

Importing "mobx-react-lite/batchingForReactDom" in an ESM bundler should import and use `observerBatching` from the same module that "mobx-react-lite" provides in the current environment.

webpack-bundle-analyzer in production:

![image](https://user-images.githubusercontent.com/6439050/88342257-b0eb5f00-cd04-11ea-8c8c-4ab318910e29.png)

## Actual outcome

Importing "mobx-react-lite/batchingForReactDom" in an ESM bundler results in duplicate code, since the CommonJS module is used by "batchingForReactDom".

webpack-bundle-analyzer in production:

![image](https://user-images.githubusercontent.com/6439050/88342396-f1e37380-cd04-11ea-9a64-0524cd3c2c1b.png)

## How to reproduce the issue

```tsx
import * as React from "react"
import * as ReactDOM from "react-dom"
import "mobx-react-lite/batchingForReactDom"
import {useObserver} from "mobx-react-lite"

const Test = () => {
  return useObserver(() => {
    return <h1>Hello World</h1>
  })
}

ReactDOM.render(<Test />, document.getElementById("app"))
```

## Versions

2.0.7

[...]

PR #300, opened 2020-08-08
Add react-dom as an optional dependency

mekwall:
Since [`batchingForReactDom.js`](https://github.com/mobxjs/mobx-react-lite/blob/45a58c42bdc4a511d55a512474fd34c4ca3afbcf/batchingForReactDom.js#L1) requires `react-dom` it should be listed as an [optional dependency](https://docs.npmjs.com/files/package.json#optionaldependencies).

This fixes the following issue when used with Yarn v2:
```
Module not found: mobx-react-lite tried to access react-dom, but it isn't declared in its dependencies; this makes the 
require call ambiguous and unsound.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,9 @@
     "mobx": "^4.0.0 || ^5.0.0",
     "react": "^16.8.0"
   },
+  "optionalDependencies": {
+    "react-dom": "^16.8.0"
+  },
   "devDependencies": {
     "@babel/core": "7.8.4",
     "@babel/preset-env": "7.8.4",


FredyC:
Why would you even use `mobx-react-lite` without React? :) And with this reasoning of yours, we would need to include `react-native` as well and only confuse people.

It doesn't matter, we are planning to change the way this works soon. Use https://github.com/mobxjs/mobx-react-lite/issues/296 for tracking when it happens.

mekwall:
@FredyC I am not using it without React and I have `react-dom` as a dependency in my project, but since Yarn v2 is enforcing [strict package boundaries](https://dev.to/arcanis/introducing-yarn-2-4eh1#strict-package-boundaries) you'll get the error mentioned unless you override the package in `.yarnrc.yml`.

FredyC:
@mekwall If you can make a PR I don't have a problem merging such a change, but include `react-native` as well.

PR #302, opened 2020-08-22
MobX 6

mweststrate:
* [x] support (and minimally require) mobx 6
* [x] kill `useAsObservableSource`, and all complexity that is introduced with it
 * Instead, write the props (or whatever) directly into the observable and
   1. use `useEffect `to update the observable after receiving new props
   2. or, to optimize away the possible double render, directly assign the observable update for render. This works fine for observables that are only used locally, but will, of course, trigger warnings when sending the observable to children, so `useEffect` approach is recommended
* [x] deprecate the second arg of `useLocalStore` (the source)
* [x] `useLocalStore` is renamed `useLocalObservable` and accept custom annotations as second arg
 * this is only a deprecation and all occurrences of `useLocalStore` should be migrated away to `useLocalObservable` to avoid statically hard to detect bugs caused by any existing second argument to be interpreted as `annotations`
* [x] deprecate `useObserver` to encourage people to use `Observer` / `observer`. This might be a bit too aggressive and probably better to export it as `_useObserver` or with a deprecation warning
* [x] deprecate the `forceUpdate` customisation option - features should be useful for basically everyone or nobody, not a single lib
* [x] Deprecate `useStaticRendering` in favor of renamed `enableStaticRendering` to distinguish from React hook.
* [ ] Drop mobx-react.js.org? as https://deploy-preview-2327--mobx-docs.netlify.app/react/react-integration.html should now have idiomatic, hook based code. What patterns are still missing there but do need documentation?

cc @FredyC @urugator @xaviergonz @olee


xaviergonz:
For mobx-react-component the pattern I used in the end is like this:

```ts
// here we define the "mobx local state" of the component
// dependencies (usually just props, but might include inputs from other hooks)
// are declared here and are turned into observables accessible through either
// this.props.NAME or this.NAME

class MyComponentState extends MobxLocalState<IMyComponentProps & { z: number }>() {
    @observable y = 0

    @computed
    get sum() {
        // x comes from component props
        // y comes from local observable
        // z comes from context
        return this.x + this.y + this.z
    }

    @action
    incY = () => {
        this.y++
    }

    getEffects() {
        // effects will be started on first render and auto disposed on unmount
        // you might use getBeforeMountEffects() instead if you want to run them
        // before the component is first mounted
        return [
            when(
                () => this.sum === 10,
                () => {
                    alert("you reached ten! (hooks / useMobxLocalState)")
                }
            ),
        ]
    }
}

// inside the component
const ctx = useContext(SomeContext)

const s = useMobxLocalState(
    // the class type
    MyComponentState,
    // here we optionally declare which vars should be "injected" and turned into an observable obj
    {
        ...props,
        ...ctx,
    },
    // the way to turn the object above into an observable is shallow by default,
    // but you can also specify "deep" or an object to cherry pick which properties
    // need to be turned into deep observables
    "shallow"
)
```

that way it is ensured that whatever props/context used by the state has been previously converted to an observable and has to be used through an object (this), rather than opening the possibility to mistakenly using destructing.

mweststrate:
Ok, cleaning up the api is probably impact full enough to justify a separate release beyond mobx 6 compatiblity: https://github.com/mobxjs/mobx-react-lite/pull/306

PR #303, opened 2020-08-25
Configure React batching automatically

mweststrate:
We used to configure batching automatically in `mobx-react`, this was made manually when `mobx-react` started to wrap `mobx-react-lite`, as it was never set up in mobx-react-lite. This PR sets up the old behavior and is largely a rip of of the react-redux configs. 

Because TSDX always rolls up the library, which makes it impossible to conditionally require a different file for native builds (since the RN package manager otherwise can pick the right file when requiring), TSDX is now only used for the UMD builds. For that reason the __DEV__ expressions are replaced with `process.env` checks, which is a transformation TSDX normally does. 

Verified:
- es builds keep source files for `reactBatchedUpdates.js` versus `reactBatchedUpdates.native.js`
- cjs (the same)
- process.env checks are transpiled away in UMD builds

Fixes:
- https://github.com/mobxjs/mobx-react/issues/859 (accidentally no longer enabling batching OOTB)
- #296 (loopback to different build format, causing the lib to be bundled twice)

I didn't test this PR in an actual project yet, so there might be stupid typo's still, as I can't publish atm :)


src/observerBatching.ts
```diff
@@ -1,25 +1,17 @@
...
-    batchedUpdates<A, B>(callback: (a: A, b: B) => any, a: A, b: B): void
-    batchedUpdates<A>(callback: (a: A) => any, a: A): void
-    batchedUpdates(callback: () => any): void
+export function defaulNoopBatch(callback: () => void) {
+    callback()
 }
 
-const observerBatchingConfiguredSymbol = getSymbol("observerBatching")
-
-export const observerBatching = (reactionScheduler?: IBatchedUpdates) => {
```
FredyC, line 13: I believe this was mentioned in readme, some people might have been using it so better to alias it to `setBatch`.

And readme should be updated as well.

FredyC (commented):
About TSDX, what about marking the `reactBatchedUpdates.ts` as external? Then it shouldn't include it and we would need such elaborate configuration here. Just a guess though.

mweststrate:
Did some tests with the mobx-react-lite@2.2.0-beta.0:

* Gets minified build in unpkg: https://unpkg.com/mobx-react-lite@2.2.0-beta.0/dist/mobxreactlite.umd.production.min.js
* Works in code-sandbox: https://codesandbox.io/s/v3v0my2370?file=/index.js
* Works in CRA fresh project
* Works in RN project

(for the latter 3 verified in debugger that a React scheduler is passed to setBatch)

mweststrate:
Got point @FredyC, I'll process them hopefully tomorrow!

mweststrate (commented):
For the TSDX updates, TSDX doesn't support the option. Rollup does support it using a custom plugin, but it means either adding a PR to TSDX, or manually wiring up rollup, typescript and babel. So kinda went the way of least resistance here :)

FredyC:
> TSDX doesn't support the option

It does support it through `tsdx.config.js` which is essentially rollup config you can tweak however you want. But as I said, I am not sure if using `external` would have a desirable effect. Up to you.

PR #306, opened 2020-08-26
[Experimental][WIP] Refactor api

mweststrate:
Follow-up experiment based on discussion in the MobX 6 branch, removing a lot of api's to keep things OSS and undo some pollution preventing this to becoming a second mobx-react:

1. kill `useAsObservableSource`, and all complexity that is introduced with it. Instead, but the props (or whatever) directly into the observable and 
   2. use `useEffect `to update the observable after receiving new props
   3. or, to optimize away the possible double render, directly assign the observable update for render. This works fine for observables that are only used locally, but will, of course, trigger warnings when sending the observable to children, so 1) is recommended
2. kill the second arg of `useLocalStore` (the source)
3. to avoid statically hard to detect bugs caused by any existing second argument to be interpreted as `annotations`, renamed it to `useLocalObservable`. But maybe we should keep it as is to simplify migration, I think that second arg is rarely used?
4. killed the `forceUpdate` customisation option. Features should be useful for basically everyone or nobody, not a single lib
5. killed `useObserver` to encourage people to use `Observer` / `observer`. This might be a bit too aggressive and probably better to export it as `_useObserver` or with a deprecation warning

I think this is such a big change, that we should release it as separate major after releasing mobx 6 compatibility, so that people don't have to migrate immediately?


mweststrate:
... although a simpler migration would probably be to still expose `useLocalStore` with deprecation message, and `useAsLocalSource` that has a deprecation message + the old naive implementation, basically:

```
useAsLocalSource(p) {
    useMemo(() => {
       console.warn("bla bla deprecated")
     })
    const o = useLocalStore(() => p)
    Object.assign(o, p)
    return o
}
```

urugator:
Imo if we keep `useLocalStore`, then people will keep using `computeds`, that depends on local vars and it will be the same discussion over and over ...
Moving local store outside the component is the way to prevent these issues. I suppose people are using `useLocalStore(initStoreFn)` anyway, so basically nothing really changes, it becomes `const [observable] = useState(initStoreFn)`.
They will simply use exactly the same tools for defining observable stores as they would anywhere else, no need to introduce anything new.
Besides `useLocalStore` becomes a one liner, so if someone for unknown reason cannot live without it and feels confident enough about himself and his colleagues, he's free to introduce it to it's codebase.

mweststrate:
I think the hook has a few benefits over the useState hook: move people away from using `setState` and forcing to use an initializer function, and auto-binding.

src/utils.ts
```diff
@@ -1,50 +1,26 @@
...
-        return window
-    }
-    if (typeof global !== "undefined") {
-        return global
-    }
-    if (typeof self !== "undefined") {
-        return self
+export function useDeprecated(msg: string) {
+    if ("production" !== process.env.NODE_ENV) {
+        useMemo(() => {
```
FredyC, line 19: And that's how people optimize early :) Honestly, I don't know what you are trying to achieve here. Do you think that lookup in a nearly empty array is slower than magic behind memorization? :)

src/useAsObservableSource.ts
```diff
@@ -0,0 +1,15 @@
+import { useDeprecated } from "./utils"
+import { observable, runInAction } from "mobx"
+import { useState } from "react"
+
+export function useAsObservableSource<TSource>(current: TSource): TSource {
+    if ("production" !== process.env.NODE_ENV)
+        useDeprecated(
+            "[mobx-react-lite] 'useAsObservableSource' is deprecated, please store the values directly in an observable, for example by using 'useObservable', and sync with 'useEffect' when needed"
```
FredyC, line 8: I suppose that's a typo? I hope you are no planning on returning `useObservable` again :)
```suggestion
            "[mobx-react-lite] 'useAsObservableSource' is deprecated, please store the values directly in an observable, for example by using 'useLocalStore', and sync with 'useEffect' when needed"
```

Anyway, I think this deprecation should probably have a link to a documentation site with a more thorough explanation.

src/useLocalStore.ts
```diff
@@ -0,0 +1,22 @@
...
+export function useLocalStore<TStore extends Record<string, any>>(initializer: () => TStore): TStore
+export function useLocalStore<TStore extends Record<string, any>, TSource extends object>(
+    initializer: (source: TSource) => TStore,
+    current: TSource
+): TStore
+export function useLocalStore<TStore extends Record<string, any>, TSource extends object>(
+    initializer: (source?: TSource) => TStore,
+    current?: TSource
+): TStore {
+    if ("production" !== process.env.NODE_ENV)
```
FredyC, line 16: Unnecessary double-check I suppose? It's done in `useDeprecated` already.

src/useLocalStore.ts
```diff
@@ -0,0 +1,22 @@
...
+        useDeprecated(
+            "[mobx-react-lite] 'useLocalStore' is deprecated, use 'useObservable' instead"
```
FredyC, line 18: Wait, what? Do you really want to re-add `useObservable`? We had that in 0.x and it was super confusing with `useObserver` ... which now doesn't exist so I guess it might be ok, but these are huge shifts in paradigm.

src/assertEnvironment.ts
```diff
@@ -1,9 +1,11 @@
 import { spy } from "mobx"
 import { useState } from "react"
 
-if (!useState) {
-    throw new Error("mobx-react-lite requires React with Hooks support")
-}
-if (!spy) {
-    throw new Error("mobx-react-lite requires mobx at least version 4 to be available")
+if ("production" !== process.env.NODE_ENV) {
```
FredyC, line 4: Why this condition? I think this is safe to have these checks in production because it would break the whole library anyway with a more cryptic error.

README.md
```diff
@@ -40,7 +36,11 @@ interface IObserverOptions {
 
 The observer converts a component into a reactive component, which tracks which observables are used automatically and re-renders the component when one of these values changes.
 
-### **`useObserver<T>(fn: () => T, baseComponentName = "observed", options?: IUseObserverOptions): T`** _([user guide](https://mobx-react.js.org/observer-hook))_
+### **`<Observer>{renderFn}</Observer>`**
+
+Is a React component, which applies observer to an anonymous region in your component.
+
+### **`useObserver<T>(fn: () => T, baseComponentName = "observed", options?: IUseObserverOptions): T`**
```
FredyC, line 43: Remove this from here since it's no longer a public API?

FredyC (commented):
I am still not convinced about removing `useObserver`. For fun, I tried [searching that name](https://github.com/search?q=useObserver&type=Code) over the GitHub, and 5000 results is not exactly a small number. And that's not considering private repos, etc... We might break a lot of codebases here without a clean way out of such breakage.

Is it really worth just to "protect" people who don't know what they are doing?

src/utils.ts
mweststrate: Wow I did this way too late in the evening. The point wasn't optimization, I was jumping between two lines of thoughts: showing it once per component (using useMemo) or once for the app (using the messages list). I think the later is the nicer thing to do.

src/useAsObservableSource.ts
mweststrate: Yes, didn't get to that yet

src/useLocalStore.ts
mweststrate: Will fix

src/useLocalStore.ts
mweststrate: Mostly to be able to arg swap the `source` with `annotations`. But maybe we can do that without renaming, as TS and runtime validation will fail if people use it wrong

src/assertEnvironment.ts
mweststrate: cause in production this check cannot have a different result as it would have in non prod (no user input involved or something)?

README.md
mweststrate: yeah didn't touch the readme.

mweststrate:
Doh, intended to rewrap `useObserver` with a deprecation message in index, but forgot about that🤦‍♂️

src/assertEnvironment.ts
FredyC: You never know, deployment is hard. It can't hurt to have it there imo.

src/utils.ts
FredyC: I see. Warning per component would be a total nightmare, but that wouldn't happen anyway here since it uses a shared scoped variable.

FredyC:
Please reconsider again deprecating `useObserver`. Did you miss my comment above?

mweststrate:
5K isn't really a big number, e.g. in comparison "observer mobx" gives 300K results. For old project it is no problem, as they use old versions of deps as well, and even with a new version it is just a single deprecated message that will only show up in DEV. I think if people use an unrecommended, soon to be undocumented api, it is no more than fair to point to them it is deprecated and how to fix it. 

Note that moving to MobX 6 will already introduce much larger changes than replacing `useObserver` with `<Observer>`, like needing to add constructors everywhere. 

Let's deprecate first, and we can always see if the fallout is too big to revert that message. But for now the balance between the support load / confusing it generates versus the unique benefits it yields don't justify its existence to me.

FredyC:
> Let's deprecate first, and we can always see if the fallout is too big to revert that message. But for now the balance between the support load / confusing it generates versus the unique benefits it yields don't justify its existence to me.

Ok, sounds fair enough.

But let's definitely keep `useLocalStore` or even a different name (useLocalObservable?). Just in case the `useObserver` stays so it doesn't clutter with `useObservable` as it was in the past.

mweststrate:
useLocalObservable sounds great!

src/useLocalStore.ts
mweststrate: Updated, left it here but removed the other one, so that the messages themselves don't end up in the prod builds.

src/assertEnvironment.ts
mweststrate: Restored

README.md
mweststrate: Updated the readme coursely

src/useLocalStore.ts
mweststrate: Renamed to `useLocalObservable` and explained the difference in readme

README.md
```diff
@@ -8,60 +8,122 @@
...
-**You need React version 16.8.0 and above**
+This is a lighter version of [mobx-react](https://github.com/mobxjs/mobx-react) which supports React **functional components only** and as such makes the library slightly faster and smaller (_only 1.5kB gzipped_). Note however that it is possible to use `<Observer>` inside the render of class components.
+Unlike `mobx-react`, it doesn't `Provider`/`inject`, as `useContext` can be used instead.
+It also doesn't offer
 
-This is a lighter version of [mobx-react](https://github.com/mobxjs/mobx-react) which supports React **functional components only** and as such makes the library slightly faster and smaller (_only 1.5kB gzipped_). In fact `mobx-react@6` has this library as a dependency and builds on top of it.
+## Compatibility table (major versions)
 
-The library does not include any Provider/inject utilities as they can be fully replaced with [React Context](https://mobx-react.js.org/recipes-context). Check out [the migration guide](https://mobx-react.js.org/recipes-migration).
+| mobx | mobx-react-lite | mobx-react<br/>(for class components) | Browser                                        |
```
FredyC, line 17: Hm, this is cool, but isn't it too confusing? I mean why to bother with `mobx-react` in here? If people want to use `mobx-react-lite` only they only need to care about MobX version, right? Some might get an impression from this table they also need to install `mobx-react` of a specific version 🤷‍♂️

README.md
```diff
@@ -8,60 +8,122 @@
...
+
+### **`useLocalObservable<T>(initializer: () => T, annotations?: AnnotationsMap<T>): T`**
+
+Creates an observable object with the given properties, methods and computed values.
+
+Note that computed values cannot directly depend on non-observable values, but only on observable values, so it might be needed to sync properties into the observable using `useEffect` (see the example below at `useAsObservableSource`).
+
+`useLocalObservable` is a short-hand for:
+
+`const state = useState(() => observable(initializer(), annotations, { autoBind: true }))[0]`
```
FredyC, line 48: That `[0]` is hard to spot, people are more used to tuple syntax imo.

```suggestion
`const [state] = useState(() => observable(initializer(), annotations, { autoBind: true }))`
```

README.md
```diff
@@ -8,60 +8,122 @@
...
+
+### **`useStaticRendering(enable: true)`**
```
FredyC, line 50: I think we should deprecate this name and find something that doesn't look like a hook :) Something like `enableStaticRendering` or even `enableSSR`.

README.md
```diff
@@ -8,60 +8,122 @@
...
 
-### **`useObserver<T>(fn: () => T, baseComponentName = "observed", options?: IUseObserverOptions): T`**
+Call `useStaticRendering(true)` when running in an SSR environment, in which `observer` wrapped components should never re-render, but cleanup after the first rendering automatically. Use `isUsingStaticRendering()` to inspect the current setting.
+
+---
+
+## Deprecated APIs
```
FredyC, line 56: Personally, I wouldn't keep this whole section here at all. Or do you actually want to keep these around in V3 and remove it in a future V4?

README.md
```diff
@@ -8,60 +8,122 @@
...
+    }, [unit])
+
+    return <h1>{state.lengthWithUnit}</h1>
+}
+```
+
+Note that, at your own risk, it is also possible to not use `useEffect`, but do `state.unit = unit` instead in the rendering.
+This is closer to the old behavior, but React will warn correctly about this if this would affect the rendering of other components.
+
+## Observer batching (deprecated)
```
FredyC, line 126: Maybe add a short explanation that this is done automatically now and they don't need to worry about it?

kubk:
To be honest, I don't understand what is the point of `useLocalStore`. Using `useState` + factory function to create a store makes more sense because it makes store logic decoupled from React & UI. Example:
```js
class Counter {
  value = 0;

  constructor() {
    makeAutoObservable(this);
  }

  increase = () => this.value++;
  decrease = () => this.value--;
}

const Component = observer(() => {
  const [counter] = useState(() => new Counter());
  ...
})
```

In this scenario, it is clear how to cover such store with unit tests. How to do it with `useLocalStore`? I also use Mobx strict mode and `useLocalStore` isn't compatible with this mode, because it doesn't convert methods to actions. It only converts methods to transactions.

FredyC:
> I also use Mobx strict mode and `useLocalStore` isn't compatible with this mode, because it doesn't convert methods to actions. It only converts methods to transactions.

This will work correctly with MobX 6

If you follow through comments here, we are going to rename `useLocalStore` to `useLocalObservable` and it will be just tiny wrapper around `observable` call with some defaults. Not everyone likes overly verbose class stores.

README.md
mweststrate: Yeah wanted to keep them around in v3, to not make migration too hard, and remove them in v4 only. Also nice to have the docs somewhere available in case people are looking at old code. So I think it is probably not too much in the way here.

README.md
mweststrate: See next line :)

README.md
mweststrate: agreed

README.md
mweststrate: yeah your right, this should probably be somewhere else :)

mweststrate:
Merging into mobx6 branch for a more holistic diff

PR #307, opened 2020-08-27
fix: Extend render props return types by `IObserverProps`

inomdzhon:


mweststrate:
Awesome, thanks! Will try to release later tonight

Issue #273, opened 2020-04-08
Naked import statement to batchingForReactDom tree-shaken in production build


JonatanRydh:
In documentation it states that you should simply add:
`import 'mobx-react-lite/batchingForReactDom';`

to configure batching. In many webpack builds however it will not work since it will be tree-shaken if it is not declared having side-effects. I guess the solution is to add: `"sideEffects": ["batching*"]` to package.json in this project.

A workaround until fixed is for users to explicit configure this like this:
```
import ReactDOM from 'react-dom';
import { observerBatching } from 'mobx-react-lite';
observerBatching(ReactDOM.unstable_batchedUpdates);
```

FredyC:
Thanks, I did not realize that. I wonder if `sideEffects` field is somewhat standardized. Will other bundlers supporting tree shaking respect it as well?

JonatanRydh:
I'm not sure how standardized it is, but at least the standard webpack configuration that comes with latest react-script will look at the current `"sideEffects":false` and happily ignore that import in production builds. So I'm thinking this should be a fairly common problem as it is now.

FredyC:
Alright, 2.0.5 is out. I tested it against the prod build of CRA and seems correct.

JonatanRydh:
Confirmed, it works in 2.0.5.

viridia:
I'm really confused about this warning, it says to import a module from mobx-react-lite, but I'm using mobx-react, which has no such import.

FredyC:
@viridia It certainly does have it since 6.2.0

PR #314, opened 2020-09-02
Remove sideEffects section, fixes #312

### LINKED ISSUE #312: Batching doesn't work with webpack in production mode in 2.2.0
## Intended outcome
Working batching.

## Actual outcome
Batching doesn't work.

## How to reproduce the issue
Try to build an application via webpack in production mode.
Minimal reproducible example: https://github.com/vkrol/mobx-react-lite-2.2.0-batching-bug (I don't know how to show this behavior in CodeSandbox). I used `patch-package` to insert `console.log` statement in `observerBatching` module, so I can check that the bundle contains this module.

I think that the problem is in the wrong `sideEffects` section in `package.json` https://github.com/mobxjs/mobx-react-lite/blob/master/package.json#L12-L14. If I remove this setting, everything works fine. I think, that we need to remove this setting completely. I can create a PR if you agree with this solution.

## Versions
`mobx-react-lite@2.2.0`.

mweststrate:
See #312. Tree shaking should already do the correct thing in mobx-react-lite if API's are unused, so there is no need to mark certain modules explicitly as (not) side effect free.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,9 +9,6 @@
   "jsnext:main": "es/index.js",
   "react-native": "es/index.js",
   "types": "dist/index.d.ts",
-  "sideEffects": [
-    "batching*"
-  ],
   "repository": {
     "type": "git",
     "url": "https://github.com/mobxjs/mobx-react-lite.git"


FredyC:
Are you absolutely sure about it? Under normal circumstances it's considered a good behavior to "announce" that the library doesn't have side effects, hence most libraries will have `sideEffects: false` in package.json. However, we clearly do have a side effect in a single file that never should be tree shaken away. That's why I believe we should keep this, just modify so it's not inside `index.ts`.

mweststrate:
Nope, I think this is the safest fix :). It is hard to verify correctness, and sideEffects is afaik a very webpack specific directive, to help it with module shaking. However, if tree shaking is applied, it doesn't matter anymore if module shaking happened (unless there are modules *with* side effects, that could still be shaken out), as tree shaking is much more effective. Also https://twitter.com/mweststrate/status/1228056670447730688 / https://twitter.com/acemarke/status/1228059504635502605. 

The risk with sideEffects field explicitly is that we have to make sure it works for the different kind of builds (cjs, es, as it depends which one is picked up by webpack etc), and might accidentally break if we decide in the future to roll those files up (as we did before < 2.2.0 and which is what most tools like TSDX / microbundle do)

FredyC:
Your call, I don't think I care that much about it to dive into it :)

Just for the reference why was this added: https://github.com/mobxjs/mobx-react-lite/issues/273

And since we still have `es` sources published, it might matter in that case 🤷

yshterev:
is this getting merged? Any temporary workaround?

mweststrate:
Merging and releasing.

PR #313, opened 2020-09-02
fix: reintroduce isObserverBatched for backwards compatibility (#309)

alex3683:
Fixes mobxjs/mobx-react-lite#309

### Code change checklist

-   [x] Added/updated unit tests
-   [ ] Updated `README` if applicable (not necessary)

PR #321, opened 2020-10-01
check ci

pustovalov:


FredyC:
Yea, it's not pick up at all, but it was working a month ago on PR #313. Something went wrong since then.

pustovalov:
@FredyC can you add some commit into this PR?

FredyC:
@pustovalov Seems you are right. No idea though how to change that, I don't see any permission settings or similar.

pustovalov:
Looks like problem was in my forked repo, I didn't add public ssh key into my circle ci repo

PR #325, opened 2020-10-11
Update link to React integration docs

nilshartmann:
The "old" link points to a page, that only contains the hint "This document has been updated and moved" and forwards to the "new" link. So I think I makes sense to directly point to the "new" URL?


FredyC:
Thanks

PR #329, opened 2020-10-18
fix: expose function to clear timers and allow immediate test exit

RoystonS:
Exposes `clearTimers` function which clears up background
'aborted render' reactions immediately.  Some test frameworks
such as Jest prevent exit unless all such timers are cleared.

For: mobxjs/mobx#2528

### Code change checklist

-   [ ] Added/updated unit tests  N/A
-   [x] Updated `README` if applicable

### DIFF
--- a/.changeset/loud-starfishes-shave.md
+++ b/.changeset/loud-starfishes-shave.md
@@ -1,0 +1,5 @@
+---
+"mobx-react-lite": patch
+---
+
+expose `clearTimers` function to tidy up background timers, allowing test frameworks such as Jest to exit immediately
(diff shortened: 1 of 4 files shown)


.changeset/loud-starfishes-shave.md
```diff
@@ -0,0 +1,5 @@
+---
+"mobx-react-lite": patch
```
FredyC, line 2:
```suggestion
"mobx-react-lite": minor
```
Technically this is minor since it adds a new public API.

FredyC:
Thanks!

PR #331, opened 2020-10-23
Remove incomplete sentence from README.md

djm:
Removes the incomplete "It also doesn't offer" sentence from the README intro.

This PR would be better solved by finishing the sentence, but I don't have the knowledge to complete what it might have been intending to say. If you let me know, I can update the PR!

Thanks for the library :)


FredyC:
Thanks 😆

PR #332, opened 2020-10-30
Use Finalization registry to dispose reactions of uncommitted components 

Bnaya:
background: https://github.com/mobxjs/mobx/discussions/2562

On this PR we introduce additional mechanism to dispose reactions from uncommitted components

The new mechanism will kick-in when platforms supports [finalization registry](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry)

High-level changes:
`reactionCleanupTracking` now expose slightly different api that is covering both mechanisms,
so `useObserver` is agnostic to the actual impl

Test changes:
The tests now must run using node 14 + `--expose-gc`
We test both impls with single jest run

### Code change checklist

-   [x] Added/updated unit tests
-   [x] Updated `README` if applicable


src/useObserver.ts
```diff
@@ -49,16 +95,14 @@ export function useObserver<T>(fn: () => T, baseComponentName: string = "observe
 
         const trackingData = createTrackingData(newReaction)
         reactionTrackingRef.current = trackingData
-        scheduleCleanupOfReactionIfLeaked(reactionTrackingRef)
```
Bnaya, line 52: In the endgame we will need to feature detect for FinalizationRegistry and use the prev impl as fallback
FinalizationRegistry is on by default in node 14, firefox, chrome, (and soonish safari?)

benjamingr:
@littledan - I don't know if this is interesting to you but I remember talking about use cases and behaviours for FinalizationRegistry some time ago (in the Node collab summit). This is a pretty solid one where React no longer guarantees it will call cleanup functions, MobX needs to run cleanup _sometime in the future_ after the state/object is no longer retained and MobX will use a FinalizationRegistry to accomplish this.

I would be curious about whether or not you think this is a good idea (it's not like we have a choice anyway 😅 🤷‍♂️ )

Bnaya:
> @littledan - I don't know if this is interesting to you but I remember talking about use cases and behaviours for FinalizationRegistry some time ago (in the Node collab summit). This is a pretty solid one where React no longer guarantees it will call cleanup functions, MobX needs to run cleanup _sometime in the future_ after the state/object is no longer retained and MobX will use a FinalizationRegistry to accomplish this.
> 
> I would be curious about whether or not you think this is a good idea (it's not like we have a choice anyway 😅 🤷‍♂️ )

Worth nothing that, this comes to replace heuristic, setTimeout based user-land garbage collector.
If the component wasn't committed after XXX time,  we will dispose the reaction.

.circleci/config.yml
```diff
@@ -21,6 +21,7 @@ jobs:
                 pkg-manager: yarn
 
             - run: yarn build
+            - run: node --version
```
Bnaya, line 24: remove

test/utils.ts
```diff
@@ -10,3 +10,17 @@ export function enableDevEnvironment() {
...
     }
 }
+
+export function sleep(time: number) {
+    return new Promise<void>(res => {
+        setTimeout(res, time)
+    })
+}
+
+export async function waitForTicks(ticks: number) {
```
Bnaya, line 20: Not use use, remove

jest.config.js
```diff
@@ -3,5 +3,10 @@ module.exports = {
     testEnvironment: "jsdom",
     setupFilesAfterEnv: [require.resolve("./jest.setup.js")],
     verbose: false,
-    coverageDirectory: "coverage"
+    coverageDirectory: "coverage",
+    globals: {
```
Bnaya, line 7: remove

package.json
```diff
@@ -23,8 +23,9 @@
         "prettier": "prettier --write \"./{src,test}/*.{js,ts,tsx}\"",
         "lint": "eslint . --ext .js,.ts,.tsx",
         "validate": "tsc --noEmit",
-        "test": "jest --watch",
-        "test:ci": "jest -i --coverage",
+        "jest": "node --expose-gc $(yarn bin jest)",
```
Bnaya, line 26: This will not work on windows machines.
Any idea what we can do there?

test/strictAndConcurrentMode.test.tsx
```diff
@@ -81,188 +75,3 @@ strictModeValues.forEach(strictMode => {
         expect(rootNode.textContent).toBe("changed")
     })
 })
-
```
Bnaya, line 84: Extracted to separate files

test/strictAndConcurrentModeUsingFinalizationRegistry.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+import * as React from "react"
+
+import { useObserver } from "../src/useObserver"
+import { sleep } from "./utils"
+import { FinalizationRegistry } from "../src/utils/FinalizationRegistryWrapper"
+
+afterEach(cleanup)
+
+test("uncommitted components should not leak observations", async () => {
+    if (!global.gc || !FinalizationRegistry) {
```
Bnaya, line 12: I don't like it, but i don't have nicer ideas

test/strictAndConcurrentModeUsingFinalizationRegistry.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+        throw new Error("This test must run with node >= 14 and --expose-gc flag")
+    }
+
+    const store = mobx.observable({ count1: 0, count2: 0 })
```
Bnaya, line 16: We may explicitly share this code part with the test with timers

package.json
FredyC: And why do we need it exactly?

Bnaya: On the Finalization registry group test we use that to force GC so the cleanup will run.
I'v tried other ways without much success.
I can give it another try

src/useObserver.ts
```diff
@@ -32,24 +33,27 @@ export function useObserver<T>(fn: () => T, baseComponentName: string = "observe
         // reaction from an abandoned render was disposed).
 
         const newReaction = new Reaction(observerComponentNameFor(baseComponentName), () => {
+            if (!reactionTrackingRef.current) {
+                // this branch is not expected to happen
```
FredyC, line 37: Maybe elaborate more on why this can never happen? I mean this whole block is into the same condition. Isn't it assumption based on the implementation of Reaction? What if at some point in the future it will be necessary to run that callback sooner?

src/useObserver.ts
```diff
@@ -20,6 +19,8 @@ export function useObserver<T>(fn: () => T, baseComponentName: string = "observe
         return fn()
     }
 
+    const [objectRetainedByReact] = React.useState({})
```
FredyC, line 22: So this is essentially an identity of component instance, that's clever :)

package.json
FredyC: Maybe this could help? https://www.npmjs.com/package/expose-gc

.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ version: 2.1
 executors:
     my-executor:
         docker:
-            - image: circleci/node:11
+            - image: circleci/node:14
```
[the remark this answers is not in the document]
Bnaya: Only node 14 (And browsers) supports un flagged finalizationregistry,
On node 12 it has the old name finalization group behind a flag

src/useObserver.ts
Bnaya: yes :)

src/useObserver.ts
Bnaya: I'm not sure if i have a good one,

Maybe I will restore the old code here, and pass the whole ref to addReactionToTrack

package.json
benjamingr: It doesn't work on windows because of the bash not because of `--expose-gc` (that works on windows).

`$(yarn bin jest)` expands to the jest location. I think since the path is already supposed to be set I would use `NODE_FLAGS` with `cross-env` (to set them up in Windows).

Alternatively, I think it's fine not to run this test on Windows.

Bnaya: NODE_FLAGS dose not work with expose gc unfortunately,
Using that package seems working, i will push it soon

src/useObserver.ts
Bnaya: Changes pushed

src/utils/reactionCleanupTracking.ts
```diff
@@ -1,4 +1,41 @@
 import { Reaction } from "mobx"
+import { FinalizationRegistry as FinalizationRegistryMaybeUndefined } from "./FinalizationRegistryWrapper"
+
+const {
+    addReactionToTrack,
+    recordReactionAsCommitted,
+    resetCleanupScheduleForTests,
+    forceCleanupTimerToRunNowForTests
+} = FinalizationRegistryMaybeUndefined
+    ? createReactionCleanupTrackingUsingFinalizationRegister(FinalizationRegistryMaybeUndefined)
```
FredyC, line 10: It would be nice to separate each of these methods into a separate file. For easier maintenance in case, we want to drop timers eventually.

package.json
```diff
@@ -1,110 +1,111 @@
 {
-    "name": "mobx-react-lite",
```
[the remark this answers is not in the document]
Bnaya: yes,
I was doing save without formatting, but missed this time
I will restore it

src/utils/reactionCleanupTracking.ts
```diff
@@ -20,9 +57,129 @@ export interface IReactionTracking {
...
+             * But we keep same api
+             */
+            objectRetainedByReact: unknown
+        ) {
+            reactionTrackingRef.current = createTrackingData(reaction)
+            scheduleCleanupOfReactionIfLeaked(reactionTrackingRef)
+            return reactionTrackingRef.current
+        },
+        recordReactionAsCommitted(reactionRef: React.MutableRefObject<IReactionTracking | null>) {
+            recordReactionAsCommitted(reactionRef)
```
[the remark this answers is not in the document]
FredyC: Oh, wait, nevermind. It's rather confusing to read. Maybe different names for either of those?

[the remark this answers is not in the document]
Bnaya: There's a function with the same name in the closure scope,
I can just put it directly in the object literal

src/utils/reactionCleanupTracking.ts
```diff
@@ -45,93 +202,49 @@ export const CLEANUP_LEAKED_REACTIONS_AFTER_MILLIS = 10_000
...
-                uncommittedReactionRefs.delete(ref)
-            }
+// FinalizationRegistry-based uncommitted reaction cleanup
+function createReactionCleanupTrackingUsingFinalizationRegister(
+    FinalizationRegistry: NonNullable<typeof FinalizationRegistryMaybeUndefined>
+): ReactionCleanupTracking {
+    const cleanupTokenToReactionTrackingMap = new Map<number, IReactionTracking>()
+    let globalCleanupTokensCounter = 1
+
+    const registry = new FinalizationRegistry(function cleanupFunction(token: number) {
```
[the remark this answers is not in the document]
Bnaya: FinalizationRegistry is passed as parameter here from the call site of createReactionCleanupTrackingUsingFinalizationRegister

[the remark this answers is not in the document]
FredyC: Right, missed that. That's fine.

mweststrate:
Didn't review the tests, but the logic itself looks solid to me! Awesome approach :)

mweststrate (approved):

src/utils/reactionCleanupTracking.ts
FredyC: @Bnaya After splitting this I think we merge :)

test/utils.ts
```diff
@@ -10,3 +10,9 @@ export function enableDevEnvironment() {
         process.env.NODE_ENV === "production"
     }
 }
+
+export function sleep(time: number) {
```
benjamingr, line 14: This can be `promise/timers` or `util.promisify(setTimeout)` :]

benjamingr (approved):
Echoing @mweststrate 's suggestion that this is useful for more than MobX react and should probably be published as a separate package as well as a generic "useCleanup" hook as well probably?

Bnaya:
I will address the comments & do more browsers testing tomorrow,
Hopefully also merged it

